PR #2, opened 2016-05-08
Fill out the list of reserved keywords.

judah:


proto-lens-protoc/src/Definitions.hs
```diff
@@ -164,6 +165,33 @@ fieldName = unpack . disambiguate . camelCase
                 [""] -> error "camelCase: name consists only of underscores"
                 s':ss -> T.concat $ underlines : toLower s' : map capitalize ss
 
+-- | A list of reserved keywords in Haskell2010.
+reservedIds :: Set.Set Text
```
blackgnezdo: Maybe rename to reservedKeywords? Does GHC expose this list by chance? Or would that bring in a large dependency?

proto-lens-tests/tests/names_test.hs
```diff
@@ -4,25 +4,115 @@
...
+    , testHaskellKeywords
+    , testProtoKeywords
+    , testProtoKeywordTypes
+    , testReadReservedName
+    ]
+
+-- | Test that we can get/set each individual field.
+testFields :: forall a . (Show a, Message a, Eq a)
+           => String -> a
+           -> [SomeLens a Int32]
```
blackgnezdo: SomeLens doesn't seem to be explicitly imported? Where did sneak in from?

proto-lens-tests/tests/names_test.hs
```diff
@@ -4,25 +4,115 @@
...
+testFields name defValue fields = testGroup name
+    [ testCase "fields" $ mapM_ testField fields
+    , runTypedTest (roundTripTest "roundTrip" :: TypedTest a)
+    ]
+  where
+    testField (SomeLens f) = 1 @=? (defValue & f .~ 1) ^. f
+
+-- | Wraps a Lens' (which is a higher-order type) so it can be used in a list
+-- without ImpredicativeTypes.
+data SomeLens a b = SomeLens (Lens' a b)
```
blackgnezdo: Would newtype not do the trick?

proto-lens-protoc/src/Definitions.hs
judah: Renamed to reservedKeywords.

Surprisingly "forall" and "hiding" are actually valid names:

    $ ghci -XScopedTypeVariables
    GHCi, version 7.10.2: http://www.haskell.org/ghc/  :? for help
    Prelude> let (forall, hiding) = (1,2)
    Prelude> (forall, hiding)
    (1,2)

  I added a link to the Haskell2010 report where I got the initial list from, and also added "mdo", "rec" and "proc" from that wiki page.

proto-lens-tests/tests/names_test.hs
judah: Done.  I was mixing up these higher-order types (where newtype works fine) for existential types (where it doesn't work).

PR #10, opened 2016-06-20
Fix sdist by pointing it to the generated files.

judah:
The generated files will get included in the output tarball.  However,
'cabal install' will still regenerate them and so require the `protoc`
executable to be installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
 
 before_install:
   - mkdir -p $HOME/.local/bin
-  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:$PATH
+  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:$HOME/.cabal/bin:$PATH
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -16,19 +17,45 @@ matrix:
...
+      cd $p;
+      SRC_TGZ=$(cabal info . | awk '{print $2 ".tar.gz";exit}');
+      cd dist;
+      if [ -f "$SRC_TGZ" ]; then
+        cabal install --force-reinstalls "$SRC_TGZ";
+        else
+          echo "expected '$SRC_TGZ' not found";
+          exit 1;
+      fi;
+      cd ..;
```
blackgnezdo: This can be avoided by nesting shells:
for p in $PACKAGES;
 +    do
 +      ( cd $p && 
 +      ...
         )

.travis.yml
```diff
@@ -16,19 +17,45 @@ matrix:
 
 before_install:
   - mkdir -p $HOME/.local/bin
-  - export PATH=$HOME/.local/bin:$PATH
+  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:$PATH
```
blackgnezdo: How does /opt come about? Do you specify it somewhere else?

judah: It's (implicitly) from the "hvr-ghc" apt repo specified in the "matrix" section's "addons" field.

PR #25, opened 2016-07-10
TextFormat.showMessage should print enum names, not numeric values.

judah:


proto-lens/src/Data/ProtoLens/TextFormat.hs
```diff
@@ -88,8 +88,7 @@ pprintField msg (name, FieldDescriptor _ typeDescr accessor)
 pprintFieldValue :: String -> FieldTypeDescriptor value -> value -> Doc
 pprintFieldValue name MessageField m
     = sep [text name <+> lbrace, nest 2 (pprintMessage m), rbrace]
--- TODO: make this output the enum value name by default
-pprintFieldValue name EnumField x = primField name $ fromEnum x
+pprintFieldValue name EnumField x = text name <> colon <+> text (showEnum x)
```
awpr: I think this will need to change again once unknown enum values support lands -- e.g. "text name <> colon <+> either (text . show) (text . showEnum) (checkEnum x)".  Nothing to do for now, though.

awpr:
LGTM.

PR #30, opened 2016-07-15
Support enum aliases using PatternSynonyms.

judah:
The first case listed in the proto for each int value is the "real" enum
constructor, and subsequent ones are synonyms.  This is the same approach as
Java:
https://developers.google.com/protocol-buffers/docs/reference/java-generated#enum

### DIFF
--- a/proto-lens-protoc/src/Definitions.hs
+++ b/proto-lens-protoc/src/Definitions.hs
@@ -232,7 +232,7 @@ enumDef protoPrefix hsPrefix d = let
             })
 
 -- | Generate the definitions for each enum value.  In particular, decide
--- whether it's a true constructor or a PatternSynonym' to another
+-- whether it's a true constructor or a PatternSynonym to another
 -- constructor.
 --
 -- Like Java, we treat the first case of each numeric value as the "real"
--- a/proto-lens-protoc/src/Definitions.hs
+++ b/proto-lens-protoc/src/Definitions.hs
@@ -14,6 +14,7 @@ module Definitions
     , MessageInfo(..)
     , FieldInfo(..)
     , EnumInfo(..)
+    , EnumValueInfo(..)
     , qualifyEnv
     , unqualifyEnv
     , collectDefinitions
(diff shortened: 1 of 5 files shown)


proto-lens-protoc/src/Definitions.hs
```diff
@@ -213,9 +228,30 @@ enumDef protoPrefix hsPrefix d = let
        , Enum EnumInfo
             { enumName = mkHsName (d ^. name)
             , enumDescriptor = d
-            , enumValueNames = fmap (mkHsName . (^. name)) (d ^. value)
+            , enumValues = collectEnumValues mkHsName $ d ^. value
             })
 
+-- | Generate the definitions for each enum value.  In particular, decide
+-- whether it's a true constructor or a PatternSynonym' to another
```
awpr: extra and/or missing single quote at "PatternSynonym"

proto-lens-protoc/src/Generate.hs
```diff
@@ -266,24 +269,41 @@ generateEnumDecls info =
...
+    constructorProtoNames = map (second (^. name)) constructors
+    constructorNumbers = map (second (fromIntegral . (^. number)))
+                          constructors
 
-    succPairs = zip ctorNames $ tail ctorNames
-      where ctorNames = map fst values
+    succPairs = zip constructorNames $ tail constructorNames
     succDecl funName boundName thePairs = InsDecl $ FunBind $
-        match funName [PApp (UnQual boundName) []] (
+        match funName [PVar boundName] (
```
awpr: Doesn't this change make all calls fail?  I think it corresponds to:

succ something = error "..."
succ Enum0 = Enum1
succ Enum1 = Enum2
...

I think the confusion is "boundName" is not a name to be bound but the name of the upper bound of the enum, so it's a data constructor to be matched on.

Although I'm not sure what will happen when you write out (PVar "EnumMax") (with a capitalized string as the variable name) -- maybe it will just write "succ EnumMax = ..." and work correctly by coincidence.

proto-lens-protoc/src/Definitions.hs
judah: Removed the single quote.

proto-lens-protoc/src/Generate.hs
judah: Reverted, sorry for the confusion.

It does work correctly either way by coincidence, but you're right that it should be PApp.

awpr:
LGTM

PR #31, opened 2016-07-17
Reexport modules with new names from proto-lens-protoc.

judah:
This should prevent user packages from accidentally depending on
the particular set of modules that we reexport.  For example,
previously they could get the module 'Data.ByteString' from
proto-lens-protoc.  After this change, they'll need to explicitly
list 'bytestring' in their 'build-depends' to get that.


awpr:
Cool, I didn't realize it was possible to do that.  LGTM.

PR #33, opened 2016-07-28
Add protoc insertion points for the imports and the end of the file.

### LINKED ISSUE #32: Generate insertion points to allow integration with other protoc plugins.
Protobuf compiler plugins can request that their generated code be pasted into an existing generated file immediately above a specified insertion point, using the File.insertion_point field.  These insertion points are specified by placing "@@protoc_insertion_point(some_name)" in the generated source.  proto-lens-protoc should have insertion points for at least the imports and the module top-level scope.

awpr:
This allows other plugins to add their own imports and declarations to the modules generated by proto-lens-protoc.

Fixes #32.

### DIFF
--- a/proto-lens-protoc/src/Generate.hs
+++ b/proto-lens-protoc/src/Generate.hs
@@ -75,8 +75,9 @@ generateModule modName imports syntaxType definitions importedEnv
               ["ScopedTypeVariables", "DataKinds", "TypeFamilies",
                "MultiParamTypeClasses", "FlexibleContexts", "FlexibleInstances",
                "PatternSynonyms"]
-              -- Allow unused imports in case we don't import anything from
-              -- Data.Text, Data.Int, etc.
+          , insertionPointPragma "pragmas"
+          -- Allow unused imports in case we don't import anything from
+          -- Data.Text, Data.Int, etc.
           , OptionsPragma noLoc (Just GHC) "-fno-warn-unused-imports"
           ]
           Nothing  -- no warning text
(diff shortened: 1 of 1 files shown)


proto-lens-protoc/src/Generate.hs
```diff
@@ -105,6 +109,25 @@ generateModule modName imports syntaxType definitions importedEnv
...
+--     _ = "@@protoc_insertion_point(something)"
+--
+-- Some other possibilities include:
+-- - a WARNING or DEPRECATED pragma
+-- - an ANN pragma for the module
+-- - a RULES pragma that will never fire (the rule name is an arbitrary string)
+-- - a no-op TH splice or quasiquoter expression
+insertionPoint :: String -> Decl
+insertionPoint nm = PatBind noLoc PWildCard
+    (UnGuardedRhs (Lit (String ("@@protoc_insertion_point("++nm++")"))))
```
judah: Spaces around the ++'s to separate them from the string constant?

awpr:
Fixed the (++) formatting.  Also added an insertion point in the module header for adding LANGUAGE / OPTIONS pragmas.

I think it may be impossible to add a (non-comment) insertion point for the export list (although luckily there currently is no export list).  So this approach will fall over if we ever start generating export lists.

An alternative approach is to re-parse the pretty-printed module into an AST annotated with SrcSpanInfo, then traverse it to generate a list of Comments forming the insertion points and re-print it with exactPrint.  This would be slower but less hacky, and wouldn't rely on insane things like OPTIONS pragmas for nonexistent tools and nameless string literal decls.  Thoughts?

judah:
If we re-parsing the file and adding comments in the middle, does that invalidate parsed source locations for declarations that happen later on in the program than the inserted comments?

I'm wondering if we're just running into fundamental limits of haskell-src-exts, and it would be simpler to give up on haskell-src-exts pretty-printing the whole file.  That is:

- change `generateModule` to return a `Text.PrettyPrint.Doc` instead of a `Module`
- use [prettyPrim](http://hackage.haskell.org/package/haskell-src-exts-1.17.1/docs/Language-Haskell-Exts-Pretty.html#v:prettyPrim) to pretty-print the individual parts of a file like `Decl` or `ExportSpec`
- Use `Text.PrettyPrint` combinators to glue the parts together, and intersperse comments with the insertion points.

Up to you; OK to punt on it if you think that would be too much work for now.

awpr:
Discussed offline -- this is getting messy and the intended use-case is better served by a separate module anyway, so I'm dropping this for now.

PR #35, opened 2016-08-11
Move some parts of protoc-gen-haskell to visible modules for use by plugins.

awpr:
This also refactors the main environment analysis a bit so it can be reused, and changes a few functions to accept/return String rather than ModuleName since the un-annotated haskell-src-exts API is on the way out.

### DIFF
--- a/proto-lens-protoc/proto-lens-protoc.cabal
+++ b/proto-lens-protoc/proto-lens-protoc.cabal
@@ -4,6 +4,9 @@ synopsis:            Protocol buffer compiler for the proto-lens library.
 description:
   Turn protocol buffer files (.proto) into Haskell files (.hs) which
   can be used with the proto-lens package.
+
+  The library component of this package contains compiler code that is not
+  guaranteed to have stable APIs.
 license:             BSD3
 license-file:        LICENSE
 author:              Judah Jacobson
@@ -14,7 +17,12 @@ build-type:          Simple
 cabal-version:       >=1.21
 
 library
-  exposed-modules:   Data.ProtoLens.Setup
+  exposed-modules:
+      Data.ProtoLens.Setup
+      Data.ProtoLens.Compiler.Combinators
+      Data.ProtoLens.Compiler.Definitions
+      Data.ProtoLens.Compiler.Generate
+      Data.ProtoLens.Compiler.Plugin
   default-language:  Haskell2010
   hs-source-dirs:    src
   build-depends:
@@ -25,6 +33,7 @@ library
       , data-default-class == 0.0.*
       , directory == 1.2.*
       , filepath == 1.4.*
+      , haskell-src-exts == 1.17.*
       , lens-family == 1.2.*
       , process >= 1.2 && < 1.5
       , proto-lens == 0.1.0.1
@@ -48,9 +57,6 @@ executable proto-lens-protoc
   other-modules:
       Bootstrap.Proto.Google.Protobuf.Compiler.Plugin
       Bootstrap.Proto.Google.Protobuf.Descriptor
-      Combinators
-      Definitions
-      Generate
 
   build-depends:
         base >= 4.8 && < 4.10
--- a/proto-lens-protoc/proto-lens-protoc.cabal
+++ b/proto-lens-protoc/proto-lens-protoc.cabal
@@ -5,8 +5,8 @@ description:
   Turn protocol buffer files (.proto) into Haskell files (.hs) which
   can be used with the proto-lens package.
 
-  The library component of this package contains compiler code that is not
-  guaranteed to have stable APIs.
+  The library component of this package contains compiler code (namely
+  Data.ProtoLens.Compiler.*) that is not guaranteed to have stable APIs.
 license:             BSD3
 license-file:        LICENSE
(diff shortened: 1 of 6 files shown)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Plugin.hs
```diff
@@ -0,0 +1,125 @@
...
+import Data.ProtoLens.Compiler.Definitions
+
+-- | The filename of an input .proto file.
+type ProtoFileName = Text
+
+data ProtoFile = ProtoFile
+  { descriptor :: FileDescriptorProto
+  , haskellModule :: ModuleName
+  , definitions :: Env Name
+  , exports :: [ProtoFileName]
```
judah: For `exports`, maybe copy some of the comments at l.104 or else just point to that function's haddocks?  (It took a minute to remind myself how this field worked with everything else.)

proto-lens-protoc/src/Data/ProtoLens/Compiler/Plugin.hs
```diff
@@ -0,0 +1,125 @@
...
+import System.FilePath (dropExtension, splitDirectories)
+
+
+import Data.ProtoLens.Compiler.Definitions
+
+-- | The filename of an input .proto file.
+type ProtoFileName = Text
+
+data ProtoFile = ProtoFile
+  { descriptor :: FileDescriptorProto
```
judah: Nit: indent 4 spaces

proto-lens-protoc/src/Data/ProtoLens/Compiler/Plugin.hs
```diff
@@ -0,0 +1,125 @@
...
+        { descriptor = f
+        , haskellModule = moduleNames ! n
+        , definitions = definitionsByName ! n
+        , exports = exportsByName ! n
+        , exportedEnv = exportedEnvs ! n
+        }
+      where
+        n = f ^. name
+
+buildEnv :: [Text] -> Map ProtoFileName ProtoFile -> Env QName
```
judah: Maybe name this something a little more specific, like `collectEnvFromDeps`?  Up to you.

Also:  pass `[ProtoFileName]` as first arg instead of `[Text]`?

proto-lens-protoc/proto-lens-protoc.cabal
```diff
@@ -4,6 +4,9 @@ synopsis:            Protocol buffer compiler for the proto-lens library.
 description:
   Turn protocol buffer files (.proto) into Haskell files (.hs) which
   can be used with the proto-lens package.
+
+  The library component of this package contains compiler code that is not
```
judah: Mention specifically `Data.ProtoLens.Compiler.*`?

proto-lens-protoc/proto-lens-protoc.cabal
```diff
@@ -14,7 +17,14 @@ build-type:          Simple
 cabal-version:       >=1.21
 
 library
-  exposed-modules:   Data.ProtoLens.Setup
+  exposed-modules:
```
judah: It's unfortunate that the current Cabal integration approach will cause any library that uses a .proto file to link against all these modules as well as haskell-src-exts.  [For example.](https://github.com/google/proto-lens/blob/master/proto-lens-tests/proto-lens-tests.cabal#L73)

Not sure if there's a better organization of libraries/modules.  What you have is fine for now.   Maybe the right solution is `custom-setup` clauses, though that's only on newer Cabal and support is somewhat spotty (e.g., https://github.com/commercialhaskell/stack/issues/2094).

proto-lens-protoc/src/Data/ProtoLens/Compiler/Plugin.hs
awpr: Added some comments.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Plugin.hs
awpr: Done

proto-lens-protoc/src/Data/ProtoLens/Compiler/Plugin.hs
awpr: Done on both counts.

proto-lens-protoc/proto-lens-protoc.cabal
awpr: Done

judah:
LGTM; I'll merge it once it passes travis.

blackgnezdo:
Is this expected?

Warning: modules not listed in proto-lens-protoc.cabal for 'proto-lens-protoc' component (add to other-modules):
    Bootstrap.Proto.Google.Protobuf.Compiler.Plugin
    Bootstrap.Proto.Google.Protobuf.Descriptor
    Data.ProtoLens.Compiler.Combinators
    Data.ProtoLens.Compiler.Definitions
    Data.ProtoLens.Compiler.Generate
    Data.ProtoLens.Compiler.Plugin
proto-lens-protoc-0.1.0.1: copy/register

PR #37, opened 2016-08-24
Bump versions to 0.1.0.2.

blackgnezdo:
Bump versions to 0.1.0.2.


judah:
The optparse and arbitrary modules haven't changed since the last release; I don't think we need to bump versions or re-release them.

judah:
LGTM

PR #39, opened 2016-08-25
import Data.Functor fmap operator on older GHCs

awpr:


blackgnezdo:
Could you get rid of 2  unused imports while you are at it? 
src/Data/ProtoLens/Setup.hs:38:1:
    The import of `Distribution.Simple.Program' is redundant
    The import of `Distribution.Simple.Program.Types' is redundant

For some reason I can't make comments inilne.

awpr:
Removed them.

blackgnezdo:
Thanks!

Issue #24, opened 2016-07-09
Add a benchmark for endoding/decoding


judah:
We should have a benchmark for encoding to/from the wire format, to make sure that our reflection and abstraction in the parser doesn't cause a significant slowdown compared to other Haskell code, and to give us more confidence when refactoring.  (So far, we haven't done any performance tuning of the code.)

One arbitrary data point: decoding a 1MB proto took ~60ms on my desktop, and decoding followed by encoding (which includes forcing all the fields) took ~230ms.  (The code was compiled with -O2.)

We should probably also benchmark the text format, though that's usually less performance-critical.

agrue:
I am working on a set of `criterion` benchmarks.

agrue:
Some updates... adding `criterion` benchmarks was quite simple. So I'm currently trying to figure out what a reasonable initial set of benchmarks looks like. The only thing remotely like a "standard benchmark" I could find was here: https://github.com/google/protobuf/tree/master/benchmarks. But I'm not sure we benefit much from imitating it; it looks like a raw speed test for a single large, heterogeneous proto. It seems more interesting to me to have a larger number of small, focused test cases to help uncover pathological cases (at least right now, with our largely unoptimized code).

So for the initial PR I guess I'll just come up with a small set of test cases that make sense to me. If anyone disagrees with the above or has any requests for specific benchmarks, please respond here. :)

judah:
That plan sounds great; thanks for taking this on!

Eventually we should test the effect of proto-lens putting all fields in a single, immutable, flat Haskell record.  For large protos my guess is it's significantly less efficient (in particular, for decoding) than languages which can do updates by mutating a single field.  The standard benchmark might help us examine that use case.

judah:
Closing since the current set of benchmarks has proven useful so far.  If we run into any particular bottlenecks in the future, we can add more specific benchmarks as needed.

PR #43, opened 2016-09-06
Add proto-lens-benchmarks package with a small set of Criterion benchmarks

agrue:
This begins to address #24. I didn't really know which benchmarks were important, so I included a small set of benchmarks that were interesting to me. It should be easy for others to add their own based on these examples and using the small BenchmarkUtil library I included. Or I am always happy to add additional benchmarks to this pull request based on review comments.

### DIFF
--- a/proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
+++ b/proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
@@ -1,0 +1,77 @@
+-- Copyright 2016 Google Inc. All Rights Reserved.
+--
+-- Use of this source code is governed by a BSD-style
+-- license that can be found in the LICENSE file or at
+-- https://developers.google.com/open-source/licenses/bsd
+
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | A utility library for writing proto-lens benchmarks.
+module Data.ProtoLens.BenchmarkUtil (protoBenchmark, benchmarkMain) where
+
+import Control.DeepSeq (NFData)
+import Criterion.Main
+import qualified Criterion.Main.Options as Criterion
+import qualified Data.ByteString as BS (length)
+import Data.Maybe (fromMaybe)
+import Data.ProtoLens
+import Options.Applicative
+
+-- | Generate a group of benchmarks for encoding and decoding the given proto
+-- message. Includes benchmarks for decoding to both weak head normal form and
+-- normal form (i.e. all fields fully forced.) There is only one encoding
+-- benchmark because protos are encoded as strict ByteStrings, so there is no
+-- difference between normal form and weak head normal form.
+protoBenchmark
+    :: forall a.
+       (NFData a, Message a)
+    => String
+    -- ^ Benchmark name.
+    -> a
+    -- ^ Protocol buffer message to encode/decode.
+    -> Benchmark
+protoBenchmark groupName proto =
+    bgroup
+        fullGroupName
+        [ bench "encode" $ nf encodeMessage proto
+        , bgroup
+              "decode"
+              [ bench "whnf" $ whnf decodeMessage' encodedProto
+              , bench "nf" $ nf decodeMessage' encodedProto]]
+  where
+    -- We must indicate to the compiler that we want to decode to the same
+    -- message type as the input proto.
+    decodeMessage' bytes = (decodeMessage bytes :: Either String a)
+    encodedProto = encodeMessage proto
+    fullGroupName =
+        groupName ++ "(" ++ prettySize (BS.length encodedProto) ++ ")"
+
+prettySize :: Int -> String
+prettySize b
+  | b < 10240 = show b ++ "B"
+  | k < 10240 = show k ++ "kB"
+  | otherwise = show m ++ "MB"
+  where
+    k = b `div` 1024
+    m = k `div` 1024
+
+-- | Main function for a benchmark suite with an optional 'size' parameter.
+benchmarkMain
+    :: Int
+    -- ^ Default benchmark size if one is not provided.
+    -> (Int -> [Benchmark])
+    -- ^ Function to generate a set of benchmarks from a provided 'size'.
+    -> IO ()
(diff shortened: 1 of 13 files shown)


stack-8.0.yaml
```diff
@@ -6,6 +6,7 @@ packages:
 - proto-lens-combinators
 - proto-lens-optparse
 - proto-lens-tests
+- proto-lens-benchmarks
```
judah: Can you also add this to the `PACKAGES` list in `travis-cabal.sh`?  (Sorry that this list is duplicated in multiple places...)

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
```diff
@@ -0,0 +1,77 @@
...
+    -- ^ Protocol buffer message to encode/decode.
+    -> Benchmark
+protoBenchmark groupName proto =
+    bgroup
+        fullGroupName
+        [ bench "encode" $ nf encodeMessage proto
+        , bgroup
+              "decode"
+              [ bench "whnf" $ whnf decodeMessage' encodedProto
+              , bench "nf" $ nf decodeMessage' encodedProto]]
```
judah: Style nit: align the last `]` with the first `[`.  For example:

    [ bench ...
    , bench ...
    ]

Also in the individual benchmark suites.

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
```diff
@@ -0,0 +1,77 @@
...
+       (NFData a, Message a)
+    => String
+    -- ^ Benchmark name.
+    -> a
+    -- ^ Protocol buffer message to encode/decode.
+    -> Benchmark
+protoBenchmark groupName proto =
+    bgroup
+        fullGroupName
+        [ bench "encode" $ nf encodeMessage proto
```
judah: I wonder if it's worth trying a version of the benchmark that doesn't go through `NFData`, instead just re-encoding the messages and using the fact that `ByteString`s are trivially in WHNF?  E.g. 

    bench "encode-decoded" $ nf (fmap encodeMessage . decodeMessage') encodedProto

Let me know whether you think that would add any value.

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
```diff
@@ -0,0 +1,77 @@
...
+        , bgroup
+              "decode"
+              [ bench "whnf" $ whnf decodeMessage' encodedProto
+              , bench "nf" $ nf decodeMessage' encodedProto]]
+  where
+    -- We must indicate to the compiler that we want to decode to the same
+    -- message type as the input proto.
+    decodeMessage' bytes = (decodeMessage bytes :: Either String a)
```
judah: Should we use `decodeMessageOrDie` instead of `decodeMessage`, so that if there's a bug in the library we'd catch it rather than reporting incorrect benchmark timings?  (I'm not sure how criterion handles exceptions.)

judah:
Thanks for doing this!  I agree we can just commit the benchmarks you have now and then add more later using the current framework.

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
agrue: Done. Is there a setting for hindent that makes it conform to our style? I've been using it on the johan-tibell setting but it does some things differently (like this). Plus I've observed it to mangle my haddocks comments on occasion...

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
agrue: You mean just to avoid the boilerplate instances for Generic and NFData? Unfortunately I don't think it adds much value, because I think we want the encoding and decoding benchmarked separately so if there's a big improvement/regression we know which side it came from. And if we have separate encoding/decoding benchmarks, there's not much point in having the combined round-trip version.

It wouldn't be much work to add them though, so if you can think of a good reason to include them I'd be happy to add them.

judah: I was thinking it might help measure the overhead of the NFData instance itself.  Not a strong preference, OK to keep as-is for now.

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
judah: Not sure; I have little experience with hindent myself.  What version are you using?  From the examples it seems like it does align the "[]":
https://github.com/chrisdone/hindent/blob/master/TESTS.md

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
agrue: Ah, good idea. Yes, we should definitely do that. I was worrying about that when writing this but totally forgot about the "OrDie" version.

I tested it out and Criterion handles exceptions quite nicely; it reports an error for the benchmark that failed.

agrue:
You're welcome! PTAL

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
agrue: That's strange. I'm using 4.6.4, which looks like the last version before 5, so hardly an out-of-date release. I must be using it wrong somehow, because as I mentioned I've seen it just trash my haddock comments in some cases, and it really doesn't seem like it should do that.

proto-lens-benchmarks/src/Data/ProtoLens/BenchmarkUtil.hs
agrue: Ah okay, that makes sense. At some point I should experiment with a benchmark to see if there's a measurable difference.

Issue #45, opened 2016-09-08
showMessage and related functions use Haskell not C string escaping conventions


mikegunter:
showMessage and the related pprintMessage and showMessageShort functions use the Haskell string escaping conventions instead of the C ones.  This means that non-printing characters get written as, e.g. "\SOH", which https://github.com/google/protobuf/blob/master/src/google/protobuf/io/tokenizer.cc#L1039 won't parse.  Worse, in Haskell the escape "\101" means decimal 101 whereas the tokenizer.cc code (following C convention) interprets that as octal, i.e. decimal 97.

mikegunter:
I have a fix in the works.

mikegunter:
Fixed by #46.

PR #47, opened 2016-09-09
Fix Issue #45 by using \x escapes for non-printing characters.

mikegunter:
Add tests.


judah:
I'm seeing a test timeout on Travis:
https://travis-ci.org/google/proto-lens/builds/158636654
Can you confirm that "stack test" works on the latest version of this PR?  If so, I'll take a closer look.

mikegunter:
The test takes a very long time but indeed passes when I have run it.  (I'm
running it under time now, so I'll be able to report how long.)

-m

mikegunter:
I bit the bullet and updated the parsing as well as formatting code to do what the distribution does.

Issue #20, opened 2016-07-09
readMessage chokes on '-delimited strings


judah:
The TextFormat encoding may use either single or double quotes (though they must match).  However proto-lens only supports double-quotes so far.

The error message looks like:
```
unexpected "'"
expecting "-", number, literal string or identifier
```

I found documentation of this behavior in the protobuf sources:
https://github.com/google/protobuf/blob/master/src/google/protobuf/io/tokenizer.h#L116

Currently proto-lens's TextFormat parser uses `Text.Parsec.Token.stringLiteral` which doesn't support single-quoted strings:
https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Token.html#v:stringLiteral

Issue #40, opened 2016-09-01
TextFormat prints fields in alphabetical order


fkm3:
Given this protobuf

``` protobuf
message Date {
  int32 year = 1;
  int32 month = 2;
  int32 day = 3;
}
```

you get

``` plain
day: 1 month: 9 year: 2016
```

I think it would be better to order by tag number.

judah:
Note: it looks like this wasn't actually fixed by PR #48; probably that reference to this issue was a typo.

fkm3:
Thanks!

PR #48, opened 2016-09-14
Fix Issues #20, #40, and #45 by using the same escaping as Protobuf d…

mikegunter:
…istribution

The Unicode escaping that the Protocol Buffer distribution supports is
not implemented for now.

Add tests.

### DIFF
--- a/proto-lens-tests/tests/text_format_test.hs
+++ b/proto-lens-tests/tests/text_format_test.hs
@@ -7,6 +7,11 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Main where
 
+import qualified Data.ByteString
+import Data.Char (ord)
+import Data.Monoid ((<>))
+import qualified Data.Text.Lazy
+import Data.Word (Word8)
 import Data.ProtoLens (
     def, Message, showMessage, showMessageShort, pprintMessage)
 import Lens.Family2 ((&), (.~))
@@ -28,6 +33,9 @@ def3 = def
 
 def4 :: Test4
 def4 = def
+
+def5 :: Test5
+def5 = def
 
 failed1 :: Maybe Test1
 failed1 = Nothing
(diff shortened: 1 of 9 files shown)


proto-lens-tests/tests/text_format_test.hs
```diff
@@ -57,8 +65,40 @@ main = testMain
...
+          -- '\o172' == '\x7a' == 'z'
+         (Just $ def2 & b .~ "\o1\o12\o123\x2\o172z3z3")
+         (Data.Text.Lazy.pack "b: \"\\001\\012\\123\\002\\172\\x7a3\\1723\"")
+    , testCase "Render string with escape sequences" $
+        escapeRendered @=? showMessageShort escapeMessage
+    , readFrom "Parse rendered string with escape sequences"
+               (Just escapeMessage) (Data.Text.Lazy.pack escapeRendered)
+    , testCase "Render bytes" $
+         invalidUTF8BytesRendered @=? showMessage invalidUTF8BytesMessage
+    , readFrom "Parse single-quote-delimited string"
```
judah: Can you add a test for an escape quote nested inside a string (assuming we don't already have it)?  For example, "abc\"def"

proto-lens/src/Data/ProtoLens/TextFormat.hs
```diff
@@ -102,11 +104,35 @@ pprintFieldValue name SFixed64Field x = primField name x
...
+-- non-ASCII characters, which depending on the locale, the distribution might
+-- not do.
+--
+-- This uses three-digit octal escapes, e.g. "\011" plus \n, \r,, \t, \', \",
+-- and \\ only.  Note that Haskell string-literal syntax calls for "\011" to be
+-- interpreted as decimal 11, rather than the decimal 9 it actually represent,
+-- so you can't use Prelude.read to parse the strings created here.
+pprintByteString :: String -> Data.ByteString.ByteString -> Doc
+pprintByteString name x = text name <> colon <+> char '\"'
+    <> text (concatMap escape $ Data.ByteString.unpack x) <> char '\"'
```
judah: Include the other ascii escapes like '\a', '\b', etc.?

mikegunter: I followed the distribution, see https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/strutil.cc#L578, in not emitting those.

This inspired me to add a test for parsing them, though.

mikegunter:
See the contents of escapeMessage and escapeRendered.

-m

PR #55, opened 2016-11-28
Add defaultMainGeneratingSpecficProtos

expipiplus1:
Also add generatingSpecficProtos.

These functions allow easily using this package to generate haskell
source for specific .proto files in Setup.hs

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
@@ -135,12 +135,8 @@ generatingSpecificProtos root getProtos hooks = hooks
     }
   where
     generateSources p l = do
-        -- `makeAbsolute f` returns the absolute path of f as though it were
-        -- specified relative to root.
-        let makeAbsolute f = if isRelative f
-                               then root </> f
-                               else f
-        files <- map makeAbsolute <$> getProtos p
+        -- Applying 'root </>' does nothing if the path is already absolute.
+        files <- map (root </>) <$> getProtos p
         generateProtos root (autogenModulesDir l) files
 
 -- | Add the autogen directory to the hs-source-dirs of all the targets in the
--- a/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
@@ -16,7 +16,9 @@
 {-# LANGUAGE CPP #-}
 module Data.ProtoLens.Setup
     ( defaultMainGeneratingProtos
+    , defaultMainGeneratingSpecificProtos
     , generatingProtos
+    , generatingSpecificProtos
     , generateProtos
     ) where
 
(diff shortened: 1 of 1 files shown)


expipiplus1:
I signed it!

judah:
With the existing code it's already possible to generate modules for a list of specific protos by enumerating them explicitly in the `.cabal` file's `extra-source-files` field.  For example:
https://github.com/google/proto-lens/blob/master/proto-lens-combinators/proto-lens-combinators.cabal#L16

Can you please explain in more detail whether that solves your desired use case, or if there's a situation it doesn't cover?

expipiplus1:
@judah The `.proto` files are not able to be listed in the `extra-source-files` field as their location is dependent on an environment variable (guaranteed to be present by our build system). 

I'm currently using my change like so:

```haskell
-- Setup.hs
main = do
  root <- getEnv "SOME_PROTO_DIR"
  defaultMainGeneratingSpecificProtos root [root </> "foo.proto"]
```

expipiplus1:
It might be worth changing this so that the specific files can be given relative to `root`.

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -77,7 +95,29 @@ defaultMainGeneratingProtos root
...
+               files
+
+-- | Augment the given 'UserHooks' to auto-generate Haskell files from the
+-- .proto files returned by a function @getProtos@.
+--
+-- Writes the generated files to the autogen directory (@dist\/build\/autogen@
+-- for Cabal, and @.stack-work\/dist\/...\/build\/autogen@ for stack).
+--
+-- Throws an exception if the @proto-lens-protoc@ executable is not on the PATH.
+generatingSpecificProtos
```
judah: I feel there's a mismatch between `generatingSpecificProtos` which takes a `PackageDescription -> IO [FilePath]` callback and `defaultMainGeneratingSpecificProtos` which just takes a `[FilePath]`.  Compare for example:

    defaultMainGeneratingSpecificProtos :: (PackageDescription -> IO [FilePath]) -> IO ()

What do you think about that API?  Or renaming one of the functions to make the asymmetry more clear?

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -77,7 +95,29 @@ defaultMainGeneratingProtos root
...
+    :: FilePath -- ^ The root directory under which .proto files can be found.
+    -> (PackageDescription -> IO [FilePath]) -- ^ A function to return a list of
+                                             -- .proto files. The cabal package
+                                             -- description is provided
```
judah: Period after `provided`.

Or alternately, something like: `Takes the Cabal package description as input.`  Up to you.

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -66,6 +68,22 @@ defaultMainGeneratingProtos
...
+-- auto-generates Haskell files from the .proto files listed. The given .proto
+-- files should be under the given root directory.
+--
+-- Writes the generated files to the autogen directory (@dist\/build\/autogen@
+-- for Cabal, and @.stack-work\/dist\/...\/build\/autogen@ for stack).
+--
+-- Throws an exception if the @proto-lens-protoc@ executable is not on the PATH.
+defaultMainGeneratingSpecificProtos
+    :: FilePath -- ^ The root directory under which .proto files can be found.
+    -> [FilePath] -- ^ The .proto files under @root@
```
judah: As you mentioned, it may make more sense for these to be relative to the root path, though I'm not sure whether that makes the code below more complicated.  Either way, let's document the behavior when they're not absolute paths.

expipiplus1:
Thanks for the feedback @judah. I've changed the behaviour when a relative path is provided and changed the type for `defaultMainGeneratingSpecificProtos` to match the other function.

judah:
It looks like your latest change broke the extra-src-files way of doing things, causing the Travis build to fail:
https://travis-ci.org/google/proto-lens/jobs/180464541#L340
Let me know if you get stuck fixing it.  (You should be able to run the tests locally with "stack test".)  Note that extra-src-files are relative to the .cabal file (as opposed to the proto root path).

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -130,7 +135,12 @@ generatingSpecificProtos root getProtos hooks = hooks
     }
   where
     generateSources p l = do
-        files <- getProtos p
+        -- `makeAbsolute f` returns the absolute path of f as though it were
+        -- specified relative to root.
+        let makeAbsolute f = if isRelative f
```
judah: I believe you can replace `makeAbsolute` with just  `</>`, which already does nothing if `f` is absolute:
https://github.com/ghc/packages-filepath/blob/master/System/FilePath/Internal.hs#L644

judah:
One minor nit, otherwise looks good!

expipiplus1:
good spot, @judah. Fixed now.

PR #56, opened 2016-11-28
Miscellaneous documentation improvements.

awpr:
While using proto-lens for a side project I took note of a few minor issues I ran into:

- The quick start instructions result in broken packages due to missing setup-depends.
- The Haddock of encodeMessage is ambiguous about which ByteString it returns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ source files from the original protocol buffer specifications (`.proto` files).
 
 First, edit the `.cabal` file of your project to:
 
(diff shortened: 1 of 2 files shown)


awpr:
As discussed offline, removed the mention of Cabal 'setup-depends' in favor of only stack 'explicit-setup-deps'.  'setup-depends' is "the right thing", but the new versions of cabal that support parsing it aren't ubiquitous yet, and 'explicit-setup-deps' alone is enough for stack.

PR #59, opened 2016-12-09
Use haskell-src-exts 1.19.0

expipiplus1:
You might not want to merge this yet until hse-1.19.0 is in stackage nightly.


judah:
Thanks for the patch.  I agree we should hold off on committing it (I think: until HSE is in an actual stackage release), especially since we don't gain any new features and arguably it makes the code a little more verbose by adding `()`s in more types and expressions.  (Maybe it could be helped by some type aliases and/or helper functions in the Combinators module.)

judah:
Closing now that #104 is out for review.

expipiplus1:
\o/

Issue #58, opened 2016-12-08
Please use vectors instead of lists for packed values


danielwaterworth:
This library is a memory hog when it comes to parsing repeated packed scalar types. It would be nice if it would use contiguous arrays instead of lazy lists where appropriate.

judah:
It may be better to just always store POD types as contiguous arrays, regardless of whether the field has the "packed" attribute.  In particular, that attribute is supposed to be both backwards- and forwards- compatible:
https://developers.google.com/protocol-buffers/docs/encoding#packed
In other words, it's problematic if a backwards-compatible change to the .proto file causes user code to break (by changing the type of the Haskell record field).

A related idea is for our codegen to create both the lenses `foo :: Lens' Foo [a]` and `foo'packed :: Lens' Foo (Unboxed.Vector a)` (for example).  Not sure whether that's worth the extra complexity vs just always providing `foo :: Lens' Foo (Unboxed.Vector a)`.

danielwaterworth:
I started implementing this and came to the same conclusion; that repeated fields should always be vectors. I have something that works, but it's still using lots of memory. I suspect I need to make things more strict. Is there a good reason not to turn on the StrictData pragma in all generated files?

judah:
It's tricky because of nested message.  Consider the following types:

    data Foo = Foo { a :: Maybe Int }
    data Bar = Bar { b :: [Foo] }

Even if you use `StrictData`, those types will be effectively lazy since `[]` and `Maybe` are lazy.

Possibly we could work around this by making the generated lens accessors force the fields, but I'm not sure that would be efficient in all use cases (for example: nested repeated messages).

Also keep in mind that the actual encoding/decoding code mainly focuses on correctness and has not been seriously optimized yet; some of your poor memory usage may be coming from that.  

Would you be able to post example code that exhibits the bad memory behavior?  Maybe see whether code that just encodes or decodes a specific message (or messages) exhibits the same behavior?

Side note: we still want to support 7.10.3 which doesn't have `-XStrictData`.  You could accomplish the same thing by annotating the field types `!` (I believe in haskell-src-exts: with `TyBang BangedTy`).

judah:
Another option, depending on how you're using the library, is to try forcing the proto data structs via `deepseq`.  Currently you need to derive the `NFData` class manually; see for example:
https://github.com/google/proto-lens/blob/master/proto-lens-benchmarks/benchmarks/Nested.hs#L24

[...]

PR #63, opened 2016-12-19
Improve space usage and time of decoding.

judah:
Motivated by #58, it turned out there were a few space leaks in the decoder:
- Made all message fields strict
- Add bang patterns to the decoding code
- Don't generate an intermediate list of TaggedValues.

These changes had the happy side effect of significantly speeding up our
decoding benchmarks.  Additionally the "decode/whnf" and "decode/nf" tests now
take the same amount of time, indicating that we do fully force the messages.

|                                              |              before  |                after |
| -------------------------------------------- | -------------------- | -------------------- |
|nested:flat(68kB)/encode                      |  3.547ms (+/- 0.009) |  2.559ms (+/- 0.003) |
|nested:flat(68kB)/decode/whnf                 | 16.304ms (+/- 0.295) |  7.225ms (+/- 0.039) |
|nested:flat(68kB)/decode/nf                   | 19.417ms (+/- 0.198) |  7.283ms (+/- 0.044) |
|nested:nested(87kB)/encode                    | 10.670ms (+/- 0.279) |  7.791ms (+/- 0.040) |
|nested:nested(87kB)/decode/whnf               | 18.310ms (+/- 0.347) | 13.749ms (+/- 0.023) |
|nested:nested(87kB)/decode/nf                 | 21.977ms (+/- 0.185) | 13.780ms (+/- 0.043) |
|int-packing:int32-unpacked(19kB)/encode       |  1.319ms (+/- 0.003) |  0.704ms (+/- 0.011) |
|int-packing:int32-unpacked(19kB)/decode/whnf  |  8.300ms (+/- 0.337) |  3.246ms (+/- 0.005) |
|int-packing:int32-unpacked(19kB)/decode/nf    |  9.772ms (+/- 0.302) |  3.269ms (+/- 0.006) |
|int-packing:int32-packed(10003B)/encode       |  0.869ms (+/- 0.003) |  0.775ms (+/- 0.002) |
|int-packing:int32-packed(10003B)/decode/whnf  |  1.809ms (+/- 0.048) |  1.670ms (+/- 0.002) |
|int-packing:int32-packed(10003B)/decode/nf    |  3.372ms (+/- 0.129) |  1.707ms (+/- 0.002) |
|unused-fields:no-unused(19kB)/encode          |  1.341ms (+/- 0.004) |  0.693ms (+/- 0.004) |
|unused-fields:no-unused(19kB)/decode/whnf     |  8.253ms (+/- 0.076) |  3.125ms (+/- 0.015) |
|unused-fields:no-unused(19kB)/decode/nf       |  9.713ms (+/- 0.072) |  3.148ms (+/- 0.016) |
|unused-fields:with-unused(19kB)/encode        |  1.363ms (+/- 0.004) |  0.720ms (+/- 0.035) |
|unused-fields:with-unused(19kB)/decode/whnf   | 12.224ms (+/- 0.096) |  3.599ms (+/- 0.099) |
|unused-fields:with-unused(19kB)/decode/nf     | 13.383ms (+/- 0.233) |  3.608ms (+/- 0.007) |

### DIFF
--- a/proto-lens/src/Data/ProtoLens/Encoding.hs
+++ b/proto-lens/src/Data/ProtoLens/Encoding.hs
@@ -8,7 +8,9 @@
 --
 -- TODO: Currently all operations are on strict ByteStrings;
 -- we should try to generalize to lazy Bytestrings as well.
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 module Data.ProtoLens.Encoding(
@@ -23,7 +25,7 @@ import Data.ProtoLens.Encoding.Bytes
 import Data.ProtoLens.Encoding.Wire
 
 import Control.Applicative ((<|>), (<$>))
-import Control.Monad (foldM)
+import Control.Monad (foldM, join)
 import Data.Attoparsec.ByteString as Parse
 import Data.Bool (bool)
 import Data.Text.Encoding (encodeUtf8, decodeUtf8')
@@ -35,6 +37,7 @@ import qualified Data.ByteString.Lazy as L
 import Data.Monoid (mconcat, mempty)
 import Data.Foldable (foldMap, toList, foldl')
 import Lens.Family2 (set, over, (^.), (&))
+import Data.Functor.Identity (Identity(..))
 
 -- TODO: We could be more incremental when parsing/encoding length-based fields,
 -- rather than forcing the whole thing.  E.g., for encoding we're doing extra
@@ -43,9 +46,41 @@ import Lens.Family2 (set, over, (^.), (&))
 -- | Decode a message from its wire format.  Returns 'Left' if the decoding
 -- fails.
 decodeMessage :: Message msg => B.ByteString -> Either String msg
-decodeMessage input =
-    parseOnly (Parse.manyTill getTaggedValue endOfInput) input
-        >>= taggedValuesToMessage
+decodeMessage input = parseOnly (parseMessage end) input
+  where
+    end = endOfInput *> pure Nothing
+
+-- | Parse a message with the given ending delimiter (which will be EndGroup in
+-- the case of a group, and end-of-input otherwise).  The ending delimiter
+-- Parser returns (Just err) if it encountered an unrecoverable error; in
+-- particular, if it found an EndGroup with an unexpected tag number.
+parseMessage :: forall msg . Message msg => Parser (Maybe String) -> Parser msg
+parseMessage end = do
+    (msg, unsetFields) <- join $ loop def requiredFields
+    if Map.null unsetFields
+        then return $ reverseRepeatedFields fields msg
+        else fail $ "Missing required fields "
+                        ++ show (map fieldDescriptorName
+                                    $ Map.elems $ unsetFields)
+  where
+    fields = fieldsByTag descriptor
+    requiredFields = Map.filter isRequired fields
+    loop :: msg -> Map.Map Tag (FieldDescriptor msg)
+            -> Parser (Parser (msg, Map.Map Tag (FieldDescriptor msg)))
+    -- Note that the following won't work:
+    -- end >>= \case {Just err -> fail err; ...}
+    -- since otherwise the below <|> will cause us to ignore an error that's
+    -- supposed to be unrecoverable.
+    loop msg unsetFields = (end >>= pure . \case
+                                Nothing -> pure (msg, unsetFields)
+                                Just err -> fail err)
+                <|> do
+                    tv@(TaggedValue tag _) <- getTaggedValue
+                    case Map.lookup (Tag tag) fields of
+                        Nothing -> loop msg unsetFields
+                        Just field -> do
+                            !msg' <- parseAndAddField msg field tv
+                            loop msg' $! Map.delete (Tag tag) unsetFields
 
 -- | Decode a message from its wire format.  Throws an error if the decoding
 -- fails.
@@ -54,30 +89,6 @@ decodeMessageOrDie bs = case decodeMessage bs of
     Left e -> error $ "decodeMessageOrDie: " ++ e
     Right x -> x
 
--- | Convert a sequence of parsed key-value pairs into a Message via its
--- descriptor. Will fail if any of the key-value pairs do not match those
--- expected by the field descriptors.
-taggedValuesToMessage :: Message msg => [TaggedValue] -> Either String msg
-taggedValuesToMessage tvs
-    | missing <- missingFields fields tvs, not $ null missing
-        = Left $ "Missing required fields " ++ show missing
-    | otherwise = reverseRepeatedFields fields <$> result
-  where
-    addTaggedValue msg tv@(TaggedValue tag _) =
-        case Map.lookup (Tag tag) fields of
-            Nothing -> return msg
-            Just field -> parseAndAddField msg field tv
-    fields = fieldsByTag descriptor
-    result = foldM addTaggedValue def tvs
-
-missingFields :: Map.Map Tag (FieldDescriptor msg) -> [TaggedValue] -> [String]
-missingFields fields
-    = map fieldDescriptorName
-        . Map.elems
-        . foldl' (\m (TaggedValue t _) -> Map.delete (Tag t) m) requiredFields
-  where
-    requiredFields = Map.filter isRequired fields
-
 runEither :: Either String a -> Parser a
 runEither (Left x) = fail x
 runEither (Right x) = return x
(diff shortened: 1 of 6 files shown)


proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -137,42 +157,51 @@ buildMessage msg = foldMap putTaggedValue (messageToTaggedValues msg)
...
+                FieldWireType wt convert _
+                    -> [TaggedValue tag $ WireValue wt
+                                            (convert src)]
+                GroupFieldType
+                    -> TaggedValue tag (WireValue StartGroup ())
+                            : messageToTaggedValues src
+                                ++ [TaggedValue tag $ WireValue EndGroup ()]
+        embedPacked src
+            = case fieldWireType typeDescriptor of
+                GroupFieldType -> fail "packged GroupFieldType"
```
awpr: "packed"

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -137,42 +157,51 @@ buildMessage msg = foldMap putTaggedValue (messageToTaggedValues msg)
...
+    let
+        embed src
+            = case fieldWireType typeDescriptor of
+                FieldWireType wt convert _
+                    -> [TaggedValue tag $ WireValue wt
+                                            (convert src)]
+                GroupFieldType
+                    -> TaggedValue tag (WireValue StartGroup ())
+                            : messageToTaggedValues src
+                                ++ [TaggedValue tag $ WireValue EndGroup ()]
```
awpr: Could be faster by changing messageFieldToVals to take a "rest" parameter (DList-style).  Maybe not worth the complexity though, since groups are deprecated.

proto-lens/src/Data/ProtoLens/Encoding.hs
awpr: What's this 'fail' intended to do?  Probably not this:

Prelude> fail "" :: [Int]
[]

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -137,42 +157,51 @@ buildMessage msg = foldMap putTaggedValue (messageToTaggedValues msg)
...
+        embedPacked src
+            = case fieldWireType typeDescriptor of
+                GroupFieldType -> fail "packged GroupFieldType"
+                FieldWireType wt convert _ -> let
+                    v = L.toStrict $ toLazyByteString
+                        $ mconcat [putWireValue wt (convert x) | x <- src]
+                    in [TaggedValue tag $ WireValue Lengthy v]
+    in case accessor of
             PlainField d f
                 | Optional <- d, src == fieldDefault -> []
```
awpr: I think this deserves a comment reminding readers that PlainField is proto3-style, and proto2-style optionals are handled below without the logic to skip encoding the default.

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -137,42 +157,51 @@ buildMessage msg = foldMap putTaggedValue (messageToTaggedValues msg)
...
-                | otherwise -> [WireValue wt (convert src)]
+                | otherwise -> embed src
               where src = msg ^. f
             OptionalField f -> case msg ^. f of
```
awpr: Does "foldMap embed (msg ^. f)" work?

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -137,42 +157,51 @@ buildMessage msg = foldMap putTaggedValue (messageToTaggedValues msg)
...
-                Just src -> [WireValue wt (convert src)]
-                _ -> mempty
+                                Just src -> embed src
+                                _ -> mempty
             RepeatedField Unpacked f
-                -> [ WireValue wt (convert src)
-                   | src <- toList (msg ^. f)
-                   ]
+                -> mconcat [embed src | src <- toList (msg ^. f)]
```
awpr: Does "foldMap embed (msg ^. f)" work?

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -199,11 +228,22 @@ fieldWireType FloatField = simpleFieldWireType Fixed32 floatToWord wordToFloat
...
+                                    (stringizeError . decodeUtf8')
 fieldWireType BytesField = identityFieldWireType Lengthy
-fieldWireType MessageField = FieldWireType Lengthy encodeMessage decodeMessage
-fieldWireType GroupField =
-    FieldWireType StartGroup messageToTaggedValues taggedValuesToMessage
+fieldWireType MessageField = FieldWireType Lengthy encodeMessage
+                                decodeMessage
+fieldWireType GroupField = GroupFieldType
+
+-- TODO: check it's the same tag, and if not then fail the *whole* computation.
```
awpr: TODONE?

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
```diff
@@ -39,19 +40,19 @@ getVarInt :: Parser Word64
 getVarInt = loop 1 0
   where
   -- TODO: bang pattern?
```
awpr: TODONE

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
```diff
@@ -73,19 +74,15 @@ getWireValue Fixed32 _ = anyBits
 getWireValue Lengthy _ = getVarInt >>= Parse.take . fromIntegral
 -- Precompute the final EndGroup tag and keep parsing key-value pairs until
 -- we reach the EndGroup.
```
awpr: Stale comment.

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
```diff
@@ -54,16 +54,17 @@ data Equal a b where
...
-equalWireTypes :: String -> WireType a -> WireType b
-               -> Either String (Equal a b)
-equalWireTypes _ VarInt VarInt = Right Equal
-equalWireTypes _ Fixed64 Fixed64 = Right Equal
-equalWireTypes _ Fixed32 Fixed32 = Right Equal
-equalWireTypes _ Lengthy Lengthy = Right Equal
-equalWireTypes _ StartGroup StartGroup = Right Equal
-equalWireTypes _ EndGroup EndGroup = Right Equal
+{-# INLINE equalWireTypes #-}
+equalWireTypes :: Monad m => String -> WireType a -> WireType b
```
judah: MonadFail was only added in ghc-8, and we still want to support 7.10 for now.

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -43,9 +46,41 @@ import Lens.Family2 (set, over, (^.), (&))
...
+  where
+    fields = fieldsByTag descriptor
+    requiredFields = Map.filter isRequired fields
+    loop :: msg -> Map.Map Tag (FieldDescriptor msg)
+            -> Parser (Parser (msg, Map.Map Tag (FieldDescriptor msg)))
+    -- Note that the following won't work:
+    -- end >>= \case {Just err -> fail err; ...}
+    -- since otherwise the below <|> will cause us to ignore an error that's
+    -- supposed to be unrecoverable.
+    loop msg unsetFields = (end >>= pure . \case
```
judah: Done, good suggestion.  It turned out that the code already fails implicitly if the EndGroup tag doesn't match; I added a unit test for this.

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -54,69 +89,54 @@ decodeMessageOrDie bs = case decodeMessage bs of
...
-missingFields :: Map.Map Tag (FieldDescriptor msg) -> [TaggedValue] -> [String]
-missingFields fields
-    = map fieldDescriptorName
-        . Map.elems
-        . foldl' (\m (TaggedValue t _) -> Map.delete (Tag t) m) requiredFields
-  where
-    requiredFields = Map.filter isRequired fields
-
 runEither :: Either String a -> Parser a
 runEither (Left x) = fail x
```
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: Punting for now.  I tried converting everything to DList and saw the benchmarks get slower.  Part of that may be that the code is already using list builder fusion which eliminates some intermediate lists.

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: Done.

judah: Replaced with "error", since that's a problem with the (generated) message descriptor rather than the input.

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: It turns out foldMap doesn't fuse the intermediate list like the previous code did.  Switched to concatMap which produces similar code to before.

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -66,14 +62,8 @@ parseMessage end = do
...
-            -> Parser (Parser (msg, Map.Map Tag (FieldDescriptor msg)))
-    -- Note that the following won't work:
-    -- end >>= \case {Just err -> fail err; ...}
-    -- since otherwise the below <|> will cause us to ignore an error that's
-    -- supposed to be unrecoverable.
-    loop msg unsetFields = (end >>= pure . \case
-                                Nothing -> pure (msg, unsetFields)
-                                Just err -> fail err)
+            -> Parser (msg, Map.Map Tag (FieldDescriptor msg))
+    loop msg unsetFields = (end >> pure (msg, unsetFields))
```
awpr: Maybe x <$ m instead of m >> pure x if you're not opposed to it.

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -234,16 +221,11 @@ fieldWireType MessageField = FieldWireType Lengthy encodeMessage
...
+endOfGroup :: String -> Int -> Parser ()
 endOfGroup name tag = do
     TaggedValue tag' (WireValue wt _) <- getTaggedValue
     Equal <- equalWireTypes name EndGroup wt
-    return $ if tag == tag'
-                then Nothing
-                else Just $ "endOfGroup: unexpected tag " ++ show tag'
-                                            ++ " for field " ++ show name
-                                                ++ ", expected " ++ show tag
+    guard (tag == tag')
```
awpr: Maybe use 'label' or whatever equivalent this library has?

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: I tried it and it doesn't affect the final error message since the outer loop `parseMessage` will always just keep going when `end` fails.  Instead, it fails in `parseAndAddField` the next time it tries to parse a tag.  This is a change from the previous logic of this PR that was able to immediately fail when this guard failed.

I think on balance the simpler code is better, especially since groups are deprecated, and we may eventually move away from attoparsec to a parser library with more limited (or no) backtracking.

Issue #38, opened 2016-08-24
In proto3, repeated fields of scalar numeric types use packed encoding by default.


pippijn:
See https://developers.google.com/protocol-buffers/docs/proto3#specifying-field-rules. proto-lens generated code does not implement this correctly. Trying to decode repeated fields of scalar numeric types (like int32 and float) results in a message like "Field 1 expects wire type 0 but found 2", because packed repeated fields are wire type 2 (see https://developers.google.com/protocol-buffers/docs/encoding#structure).

iphydf:
Also, "Protocol buffer parsers must be able to parse repeated fields that were compiled as packed as if they were not packed, and vice versa. This permits adding [packed=true] to existing fields in a forward- and backward-compatible way."

judah:
Fixed by #64.

PR #64, opened 2016-12-28
Fix the handling of packed repeated fields (#38).

judah:
- Don't emit empty packed fields during encoding.
- Allow parsing packed fields as unpacked and vice versa.
- Proto3 repeated fields are packed by default.

### DIFF
--- a/proto-lens/src/Data/ProtoLens/Encoding.hs
+++ b/proto-lens/src/Data/ProtoLens/Encoding.hs
@@ -101,7 +101,7 @@ parseAndAddField
           getPackedVals = case fieldWireType typeDescriptor of
             GroupFieldType -> fail "Groups can't be packed"
             FieldWireType fieldWt _ get -> runEither $ do
-              Equal <- equalWireTypes name wt Lengthy
+              Equal <- equalWireTypes name Lengthy wt
               let getElt = do
                         wv <- getWireValue fieldWt tag
                         x <- runEither $ get wv
--- a/proto-lens/src/Data/ProtoLens/Encoding.hs
+++ b/proto-lens/src/Data/ProtoLens/Encoding.hs
@@ -100,7 +100,7 @@ parseAndAddField
           -- Get a block of packed values, reversed.
           getPackedVals = case fieldWireType typeDescriptor of
             GroupFieldType -> fail "Groups can't be packed"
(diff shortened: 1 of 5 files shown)


proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -89,15 +89,15 @@ parseAndAddField :: msg
...
-    (TaggedValue tag (WireValue wt val))
+    (TaggedValue tag (WireValue (wt :: WireType w) val))
     = case fieldWireType typeDescriptor of
-        FieldWireType fieldWt _ get -> let
+        (FieldWireType fieldWt _ get :: FieldWireType a) -> let
           getSimpleVal = do
               Equal <- equalWireTypes name fieldWt wt
               get val
           -- Get a block of packed values, reversed.
+          getPackedVals :: w ~ B.ByteString => Either String [a]
```
awpr: Would this be simpler as "getPackedVals :: B.ByteString -> Either String [a]" and then the type equality can be implicit at the call site (in the RepeatedField case below)?

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -107,12 +107,20 @@ parseAndAddField
...
-                  return $ over f (x:) msg
-              RepeatedField Packed f -> do
-                  xs <- getPackedVals
-                  return $ over f (xs++) msg
+              RepeatedField _ f
+                    -- Parse either a packed or unpacked representation,
+                    -- depending on how it was encoded.
+                    -- Note that if fieldWt is Lengthy (e.g., "string" or
+                    -- message) we should always parse it as unpacked.
+                    | Right Equal <- equalWireTypes name wt fieldWt -> do
```
awpr: Alternatively (ha), "RepeatedField _ f -> getPackedVals <|> getUnpackedVals <|> fail ..." and keep the "equalWireTypes" calls in the helper functions.

judah: Done.

proto-lens/src/Data/ProtoLens/Encoding.hs
judah: It turned out GHC can infer these types (maybe due to the refactor you suggested below), so I just removed the type signature.

PR #65, opened 2016-12-30
Support OverloadedLabels.

judah:
Adds a new package `lens-labels` (replacing Data.ProtoLens.Field)
which supports a newtype'd version of lenses, along with a typeclass
`HasLens` for using them with OverloadedLabels.

Note: In some cases it seems this causes problems with the lens-family
package and GHC's RankNTypes type inference, it seems because lens-family makes
setter functions like (.~) into higher-rank functions than are strictly
necessary.  Compare for example:
- http://hackage.haskell.org/package/lens-family-1.2.1/docs/Lens-Family2.html#v:.-126-
- http://hackage.haskell.org/package/lens-family-core-1.2.1/docs/Lens-Family.html#v:.-126-
Luckily, we can work around this by switching to a different package such as
`lens-family-core`, `lens` or `microlens`.


lens-labels/src/Lens/Labels.hs
```diff
@@ -0,0 +1,106 @@
...
+import Data.Functor.Const
+#endif
+import Data.Functor.Identity
+
+
+-- | A newtype for defining lenses.  Can be composed using
+-- `(Control.Category..)` (also exported from this module).
+newtype (~>) a b = LensFn {runLens :: a -> b}
+
+type LensFn = (~>)
```
blackgnezdo: Is ~> prior art or is it irresistibly beautiful? :)
I'm just not sure if this is easy enough to parse compared to the more boring LensFn.

proto-lens-tests/tests/labels_test.hs
```diff
@@ -0,0 +1,59 @@
...
+    let msg3 = def & runLens #a Lens.Family2..~ 42 :: Test1
+    msg1 @?= msg2
+    msg1 @?= msg3
+
+-- | lens-family-core doesn't need any extra annotations to infer the type
+-- of `msg2`.
+lensFamilyCore :: IO ()
+lensFamilyCore = let
+    msg1 = def & #a .~ 42 :: Test1
+    msg2 = def & runLens #a Lens.Family..~ 42
```
blackgnezdo: ..~ is pretty wild syntactically :)

lens-labels/src/Lens/Labels.hs
judah: Removed in favor of `LensFn`.

Most practical code will use `LensLike`, `Lens` or another type synonym, so `~>` would only appear in type errors, where I'm on the fence about it helping, so it's probably better to be boring as you suggested.

proto-lens-tests/tests/labels_test.hs
judah: Switched to the non-operator `set` which does probably make the test easier to parse.

bootstrap.hs
```diff
@@ -22,7 +22,9 @@ main = wrapClean $ do
   [installRoot] <- lines <$> readProcess "stack"
                     [useBootstrappingYaml, "path", "--local-install-root"] ""
   let protocGenHaskell = installRoot </> "bin/proto-lens-protoc"
-  callProcess "stack" [useBootstrappingYaml, "build", "proto-lens-protoc"]
+  callProcess "stack" [useBootstrappingYaml, "build", "proto-lens-protoc",
```
blackgnezdo: Should this be a separate PR?

lens-labels/lens-labels.cabal
```diff
@@ -0,0 +1,33 @@
...
+category:            Data
+build-type:          Simple
+cabal-version:       >=1.10
+
+library
+  exposed-modules:     Lens.Labels
+  build-depends:       base >= 4.8 && < 4.10
+                     , ghc-prim >=0.4 && <0.6
+  default-extensions:
+        CPP
```
blackgnezdo: Would it be better to stick these into the .hs on the theory that they are the property of the source file (also more precise when different files need different extensions)?

lens-labels/src/Lens/Labels.hs
```diff
@@ -0,0 +1,104 @@
...
+import qualified Control.Category as Category
+import GHC.Prim (Proxy#, proxy#)
+#if __GLASGOW_HASKELL__ >= 800
+import GHC.OverloadedLabels (IsLabel(..))
+#endif
+import GHC.TypeLits (Symbol)
+
+import Data.Function ((&))
+#if __GLASGOW_HASKELL__ >= 800
+import Data.Functor.Const
```
blackgnezdo: Enumerate imports also for Identity?

lens-labels/lens-labels.cabal
judah: Done.

lens-labels/src/Lens/Labels.hs
judah: Done.

bootstrap.hs
judah: Reverted; it was an artifact of a merge with the previous PR #68.

lens-labels/src/Lens/Labels.hs
```diff
@@ -0,0 +1,114 @@
+{- | A lens library that integrates with OverloadedLabels.
+
+Unlike the `lens` package (and others), lenses are defined as a newtype
+instead of a type synonym, to avoid overlapping with other OverloadedLabel
```
awpr: "OverloadedLabels" or "IsLabel"

lens-labels/src/Lens/Labels.hs
```diff
@@ -0,0 +1,114 @@
...
+import Data.Functor.Const(Const(..))
+#endif
+import Data.Functor.Identity(Identity(..))
+
+
+-- | A newtype for defining lenses.  Can be composed using
+-- `(Control.Category..)` (also exported from this module).
+newtype LensFn a b = LensFn {runLens :: a -> b}
+
+instance Category.Category LensFn where
```
awpr: I wonder if this can be derived by GeneralizedNewtypeDeriving?

lens-labels/src/Lens/Labels.hs
```diff
@@ -0,0 +1,114 @@
...
+type Lens s t a b = forall f . Functor f => LensLike f s t a b
+
+-- | A type class for lens fields.
+class HasLens (x :: Symbol) f s t a b
+        | x s -> a, x t -> b, x s b -> t, x t a -> s where
+    lensOf :: Proxy# x -> (a -> f b) -> s -> f t
+
+#if __GLASGOW_HASKELL__ >= 800
+instance
+    (p ~ (a -> f b), HasLens x f s t a b)
```
awpr: Could you comment as to why this type-equality trick is needed for the first argument but not the second?

lens-labels/src/Lens/Labels.hs
```diff
@@ -0,0 +1,114 @@
...
+type Getting r s t a b = LensLike (Const r) s t a b
+
+(^.), view :: s -> Getting a s t a b -> a
+s ^. f = getConst $ runLens f Const s
+view = (^.)
+
+infixl 8 ^.
+
+#if __GLASGOW_HASKELL__ < 800
+newtype Const a b = Const { getConst :: a }
```
awpr: Looks like the Const functor is in Control.Applicative in older versions of base.

lens-labels/src/Lens/Labels.hs
```diff
@@ -0,0 +1,114 @@
...
+    lensOf :: Proxy# x -> (a -> f b) -> s -> f t
+
+#if __GLASGOW_HASKELL__ >= 800
+instance
+    (p ~ (a -> f b), HasLens x f s t a b)
+    => IsLabel x (LensFn p (s -> f t)) where
+    fromLabel p = LensFn $ lensOf p
+#endif
+
+type ASetter s t a b = LensLike Identity s t a b
```
awpr: Would it be better to wrap the interface of an existing lens library with appropriate 'coerce' calls instead of implementing a subset of the functions?

lens-labels/src/Lens/Labels.hs
judah: Done.

lens-labels/src/Lens/Labels.hs
judah: Done.

lens-labels/src/Lens/Labels.hs
judah: Switched to use type equality for both arguments.

Eventually we might want to make this more flexible to support more types of lenses (prisms, indexed).  But for now I think it's more helpful for this instance to capture all possible uses of IsLabel.

lens-labels/src/Lens/Labels.hs
judah: It's tricky because the functions are higher-order which doesn't interact well with 'coerce'.  And the definitions of these functions are simple enough (and/or defined in terms of each other, e.g., `.~` and `%~`.

lens-labels/src/Lens/Labels.hs
judah: Done.

PR #66, opened 2017-01-02
Make the bootstrapping process more robust:

judah:
- Make a new package 'proto-lens-descriptors' for the descriptors
- Name those modules "Proto.*" instead of "Bootstrap.Proto.*"
- Add special logic so the descriptors modules don't use the reexported
  modules from proto-lens-protoc (otherwise we'd get a circular dependency)
- Make the bootstrapping script use a special stack YAML file.

### DIFF
--- a/proto-lens-descriptors/LICENSE
+++ b/proto-lens-descriptors/LICENSE
@@ -1,0 +1,30 @@
+Copyright Author name here (c) 2017
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+    * Redistributions in binary form must reproduce the above
+      copyright notice, this list of conditions and the following
+      disclaimer in the documentation and/or other materials provided
+      with the distribution.
+
+    * Neither the name of Author name here nor the names of other
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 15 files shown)


proto-lens-descriptors/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright Author name here (c) 2017
```
blackgnezdo: Update?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -60,15 +61,21 @@ fileSyntaxType f = case f ^. syntax of
     "" -> Proto2  -- The proto compiler doesn't set syntax for proto2 files.
     s -> error $ "Unknown syntax type " ++ show s
 
+-- Whether to import the "Reexport" modules or the originals;
+-- e.g., Data.ProtoLens.Reexport.Data.Map vs Data.Map.
+data UseReexport = UseReexport | UseOriginal
```
blackgnezdo: Maybe change this to [newtype around] ModuleName -> ImportDecl? UseOriginal would become `importSimple`.

judah: Switched to `ImportDecl -> ImportDecl`.  Since that type's pretty simple, I don't think it's worth adding a newtype.

proto-lens-descriptors/LICENSE
judah: Done.

judah:
> Does this get proto-lens closer to being completely buildable by cabal?
You mean for the Hackage releases to not require `protoc`?  Maybe a little, I'm not sure.  I think the hard part will be to work around some issues in how Cabal unpacks tarballs; see for example:
https://github.com/haskell/cabal/pull/2362
https://github.com/haskell/cabal/issues/2311

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -71,11 +72,11 @@ data UseReexport = UseReexport | UseOriginal
 generateModule :: ModuleName
                -> [ModuleName]  -- ^ The imported modules
                -> SyntaxType
-               -> UseReexport
+               -> (ImportDecl -> ImportDecl)
```
blackgnezdo: A type synonym would still be nice for discoverability. You already found a name for it: ModifyImports.

judah: Done.

PR #68, opened 2017-01-05
Make bootstrapping more robust:

judah:
- Completely hide all library fields of proto-lens-protoc when only building
  the executable.  In particular, the `build-depends` packages may change
  between versions of the library, and otherwise stack would still try to
  resolve them.
- Do a full clean before and after running the bootstrapping script.  Stack
  doesn't seem robust enough to handle all edge cases, and I frequently had to
  do manual 'stack clean's in order to get things to work.

### DIFF
--- a/proto-lens-protoc/proto-lens-protoc.cabal
+++ b/proto-lens-protoc/proto-lens-protoc.cabal
@@ -36,7 +36,7 @@ library
     default-language:  Haskell2010
     hs-source-dirs:    src
     build-depends:
-            Cabal >= 1.22 && < 1.25
+          Cabal >= 1.22 && < 1.25
         , base >= 4.8 && < 4.10
         , bytestring == 0.10.*
         , containers == 0.5.*
@@ -54,7 +54,7 @@ library
         -- For forwards compatibility, reexport them as new module names so that
         -- other packages don't accidentally write non-generated code that
         -- relies on these modules being reexported by proto-lens-protoc.
-            Data.ByteString as Data.ProtoLens.Reexport.Data.ByteString
+          Data.ByteString as Data.ProtoLens.Reexport.Data.ByteString
         , Data.Default.Class as Data.ProtoLens.Reexport.Data.Default.Class
         , Data.Map as Data.ProtoLens.Reexport.Data.Map
         , Data.ProtoLens as Data.ProtoLens.Reexport.Data.ProtoLens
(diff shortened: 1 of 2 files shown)


proto-lens-protoc/proto-lens-protoc.cabal
```diff
@@ -26,42 +26,43 @@ flag only-executable
...
+    exposed-modules:
+        Data.ProtoLens.Compiler.Combinators
+        Data.ProtoLens.Compiler.Definitions
+        Data.ProtoLens.Compiler.Generate
+        Data.ProtoLens.Compiler.Plugin
+        Data.ProtoLens.Setup
+    default-language:  Haskell2010
+    hs-source-dirs:    src
+    build-depends:
+            Cabal >= 1.22 && < 1.25
```
blackgnezdo: Strange indent, also below reexported-modules.

proto-lens-protoc/proto-lens-protoc.cabal
```diff
@@ -26,42 +26,43 @@ flag only-executable
...
-      , Data.Default.Class as Data.ProtoLens.Reexport.Data.Default.Class
-      , Data.Map as Data.ProtoLens.Reexport.Data.Map
-      , Data.ProtoLens as Data.ProtoLens.Reexport.Data.ProtoLens
-      , Data.ProtoLens.Message.Enum as Data.ProtoLens.Reexport.Data.ProtoLens.Message.Enum
-      , Data.Text as Data.ProtoLens.Reexport.Data.Text
-      , Lens.Family2 as Data.ProtoLens.Reexport.Lens.Family2
-      , Lens.Family2.Unchecked as Data.ProtoLens.Reexport.Lens.Family2.Unchecked
+  } else {
+    exposed-modules:
+        Data.ProtoLens.Compiler.Combinators
```
blackgnezdo: Does Cabal require commas in multi-line lists like build-depends below and not here?

judah: Yeah, it looks like `reexported-modules` and `build-depends` need the commas but `exposed-modules` doesn't.  Maybe because in the former, each entry can consist of multiple words.

proto-lens-protoc/proto-lens-protoc.cabal
judah: Done.

PR #76, opened 2017-03-22
Use the "Annotated" API from haskell-src-exts-1.17.*.

judah:
Adapted from PR #59 by @expipiplus1.

This is the first step to building `proto-lens` with both older and newer versions of
`haskell-src-exts`, and thus thus getting proto-lens into stackage nightly.

This switches to `haskell-src-exts<=1.17`'s `Language.Haskell.Exts.Annotated.*`
API, which corresponds to `haskell-src-exts>=1.18`'s
`Language.Haskell.Exts.*` .  Annoyingly, every datatype in the new API is
parametrized by a type, which is just `()` in our case.  To simplify the
codegen modules, I moved more functionality into the `Combinators` module and exposed
an API similar to what we had before.

Along the way, I also fixed a bug in the codegenerated error messages.  I
re-ran the bootstrap script to make sure no other changes were introduced.

The next PR will wrap the imports in CPP (exposing a common API from
`Combinators`) so we can support both versions of haskell-src-exts (and in
particular, multiple Stackage snapshots).

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
@@ -5,19 +5,205 @@
 -- https://developers.google.com/open-source/licenses/bsd
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
--- | Some utility functions, classes and instances for nicer code generation
--- with haskell-src-exts.
+{-# LANGUAGE FlexibleInstances #-}
+-- | Some utility functions, classes and instances for nicer code generation.
 --
--- In particular, we define orphan instances of IsString for various syntax
+-- Re-exports simpler versions of the types and constructors from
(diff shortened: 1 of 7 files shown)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
```diff
@@ -5,19 +5,205 @@
...
+
+type Module = Syntax.Module ()
+
+module' :: ModuleName -> [ModulePragma] -> [Syntax.ImportDecl ()] -> [Decl] -> Module
+module' modName
+    = Syntax.Module ()
+        (Just $ Syntax.ModuleHead () modName
+                    -- no warning text
+                    Nothing
+                    -- no explicit exporst; we export everything.
```
blackgnezdo: exports

expipiplus1:
Thanks, @judah!

PR #81, opened 2017-05-22
Allow .proto files to import between Haskell packages.

judah:
Also adds the `proto-lens-protobuf-types` package which includes
`Proto.Google.Protobuf.{Any,Duration,Wrappers}`.

This change allows one Haskell package to generate bindings for `foo.proto`,
and another package to generate bindings for `bar.proto`, where
`bar.proto` imports `foo.proto`.

Under the hood, each package copies its `.proto` files into
`$datadir/proto-lens-imports`, so that other packages can find
and use them when compiling their own `.proto` files.

Note: a minor corner case is not handled correctly: if a test or
benchmark requires va .proto that imports from another package,
the *library* also needs to specify that dependency.  See
`proto-lens-tests.cabal` for more details.

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
@@ -19,6 +19,7 @@ module Data.ProtoLens.Setup
     , defaultMainGeneratingSpecificProtos
     , generatingProtos
     , generatingSpecificProtos
+    , generateProtosWithImports
     , generateProtos
     ) where
 
@@ -26,6 +27,8 @@ module Data.ProtoLens.Setup
 import Data.Functor ((<$>))
 #endif
 
+import Control.Monad (filterM, forM_, when)
+import qualified Distribution.InstalledPackageInfo as InstalledPackageInfo
 import Distribution.PackageDescription
     ( PackageDescription(..)
     , benchmarkBuildInfo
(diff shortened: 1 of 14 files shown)


judah:
@awpr, interested in taking a look?

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -49,9 +63,16 @@ import System.FilePath
...
+    )
+import System.Directory
+    ( createDirectoryIfMissing
+    , doesDirectoryExist
+    , findExecutable
+    , removeDirectoryRecursive
     )
-import System.Directory (createDirectoryIfMissing, findExecutable)
 import System.Process (callProcess)
+import Distribution.Verbosity (Verbosity)
```
blackgnezdo: Sort?

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -123,21 +144,67 @@ generatingSpecificProtos
...
+    { buildHook = \p l h f -> generate p l >> buildHook hooks p l h f
+    , haddockHook = \p l h f -> generate p l >> haddockHook hooks p l h f
+    , replHook = \p l h f args -> generate p l >> replHook hooks p l h f args
     , sDistHook = \p maybe_l h f -> case maybe_l of
             Nothing -> error "Can't run protoc; run 'cabal configure' first."
             Just l -> do
-                        generateSources p l
+                        generate p l
                         sDistHook hooks (fudgePackageDesc l p) maybe_l h f
+    , postCopy = \a flags pkg lbi -> do
```
blackgnezdo: Out of curiosity, why this style over
postCopy a flags pkg lbi = do
?

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -123,21 +144,67 @@ generatingSpecificProtos
...
+generateSources :: FilePath -- ^ The root directory
+                -> LocalBuildInfo
+                -> [FilePath] -- ^ Proto files relative to the root directory.
+                -> IO ()
+generateSources root l files = do
+    -- Collect import paths from build-depends of this package.
+    importDirs <- filterM doesDirectoryExist
+                     [ InstalledPackageInfo.dataDir info </> protoLensImportsPrefix
+                     | (_, c, _) <- componentsConfigs l
+                     , (_, i ) <- componentPackageDeps c
```
blackgnezdo: Remove space in i) ?

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
judah: Done.

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
judah: Haskell's syntax doesn't let you do that when defining fields in a record :(

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
judah: Done.

PR #83, opened 2017-05-31
Use custom-setup in the packages and Readme.

judah:
Stack v1.4.0 honors it even with older versions of Cabal.  So it's
fine to just use everywhere.


judah:
PTAL; I've merged the committed version of #81 into this one and fixed conflicts.

(I don't have write access to merge this change.)

PR #87, opened 2017-06-23
Bump proto-lens-arbitrary to support newer QuickCheck.

judah:
Tested by manually editing `stack.yaml`, specifying the new versions in its `extra-deps`.  All tests passed.


blackgnezdo:
LGTM

PR #88, opened 2017-07-12
Add Data.ProtoLens.Any for storing arbitrary Messages.

### LINKED ISSUE #22: Support "Any"
We should support the `Any` type that was introduced in proto3:
https://developers.google.com/protocol-buffers/docs/proto3#any

At its core, `Any` is just another protocol buffer message (defined in `google/protobuf/any.proto`), so we should already be able to handle protos that reference it.  However, we can add a nicer API on top for converting to/from an arbitrary message type, siimilar to what the C++ and Java bindings provide.

judah:
Fixes #22.

Provides simple `pack` and `unpack` functions.  Since the `unpack` operation
is partial (and has multiple types of "failure"), I wasn't sure how to fit it
into a `Lens`; so I left that alone for the time being.

I turned this into a minor version bump.  It does add a new field to
`MessageDescriptor` which usually requires a major bump; but since that field
is internal to codegen, it shouldn't break any individual users.

### DIFF
--- a/proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
+++ b/proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
@@ -10,7 +10,8 @@ module Data.ProtoLens.Any
 
 import Control.Exception (Exception(..))
 import Data.Monoid ((<>))
-import Data.Text (Text, breakOnEnd)
+import qualified Data.Text as Text
+import Data.Text (Text)
 import Data.Typeable (Typeable)
 import Data.ProtoLens
     ( decodeMessage
@@ -30,7 +31,6 @@ pack = packWithPrefix googleApisPrefix
 googleApisPrefix :: Text
 googleApisPrefix = "type.googleapis.com"
 
-
 -- | Packs the given message into an 'Any' using the given type URL prefix.
 packWithPrefix :: forall a . Message a => Text -> a -> Any
 packWithPrefix prefix x =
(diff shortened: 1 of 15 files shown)


proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
```diff
@@ -0,0 +1,68 @@
...
+
+-- | Packs the given message into an 'Any' using the default type URL prefix
+-- "type.googleapis.com".
+pack :: forall a . Message a => a -> Any
+pack = packWithPrefix googleApisPrefix
+
+googleApisPrefix :: Text
+googleApisPrefix = "type.googleapis.com"
+
+
```
blackgnezdo: Drop empty line? Also in L41.

proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
```diff
@@ -0,0 +1,68 @@
...
+-- | Packs the given message into an 'Any' using the given type URL prefix.
+packWithPrefix :: forall a . Message a => Text -> a -> Any
+packWithPrefix prefix x =
+    def & typeUrl .~ (prefix <> "/" <> name)
+        & value .~ encodeMessage x
+  where
+    name = messageName (descriptor :: MessageDescriptor a)
+
+
+data UnpackError
```
blackgnezdo: Maybe type level haddock?

proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
```diff
@@ -0,0 +1,68 @@
...
+    = DifferentType
+        { expectedMessageType :: Text -- ^ The expected @packagename.messagename@
+        , actualUrl :: Text -- ^ The typeUrl in the 'Any' being unpacked
+        }
+    | DecodingError String  -- ^ The error from decodeMessage
```
blackgnezdo: Seems counter intuitive to have String and Text fields in the same type. Comment why it makes sense?

proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
judah: Done.

proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
judah: Done.

proto-lens-protobuf-types/src/Data/ProtoLens/Any.hs
judah: Switched to Text.

I originally used String since that's what we get from `decodeMessage :: Message a => Either String a`.  But since we have a new type, there isn't a strong need for them to be consistent.

PR #91, opened 2017-07-16
Print a better error message when missing protoc/proto-lens-protoc.

judah:
Example output when `protoc` is missing:

    Configuring proto-lens-protobuf-types-0.2.1.0...
    proto-lens-protobuf-types-0.2.1.0: build (lib)
    ==========
    Error: couldn't find the executable "protoc" in your $PATH.
	Follow the installation instructions at https://google.github.io/proto-lens/installing-protoc.html .
    ==========
    Missing executable "protoc"
    CallStack (from HasCallStack):
    error, called at src/Data/ProtoLens/Setup.hs:296:13 in
    proto-lens-protoc-0.2.1.0-EdtYegZINjJ3cgy3Ej8Sfv:Data.ProtoLens.Setup

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Setup.hs
@@ -72,6 +72,7 @@ import System.Directory
     , findExecutable
     , removeDirectoryRecursive
     )
+import System.IO (hPutStrLn, stderr)
 import System.Process (callProcess)
 
 -- | This behaves the same as 'Distribution.Simple.defaultMain', but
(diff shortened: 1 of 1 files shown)


proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -263,14 +264,33 @@ generateProtosWithImports
...
+-- found.
+findExecutableOrDie :: String -> String -> IO FilePath
+findExecutableOrDie name debugMsg = do
+    maybePath <- findExecutable name
+    case maybePath of
+        Just path -> return path
+        Nothing -> do
+            let sep = "=========="
+            hPutStrLn stderr sep
+            hPutStrLn stderr $ "Error: couldn't find the executable " ++ show name
```
blackgnezdo: Error diagnostics in the change description says "proto", I suspect it should be "protoc".

judah: Good catch, fixed.  (It was only the change description that had the error.)

PR #85, opened 2017-06-13
Sum types for oneof fields

gseitz:
This uses an intermediate sum type encoding for `oneof` fields. The generated lenses don't expose the sum type but rather point directly to (`Maybe`) the underlying type.

See (#21).

PR #92, opened 2017-07-19
Add a lens for internal "oneof" sum types.

### LINKED ISSUE #21: Better support for "oneof" fields
We currently treat "oneof" fields similar to optional fields.  This is an intended backwards-compatible behavior of the wire encoding, but makes some use cases more awkward.

https://developers.google.com/protocol-buffers/docs/proto#oneof

One possible approach is to store the value as a sum type internally, and provide lenses that return a default value when their case isn't set (as well as "maybe'foo" variants).  Another option (less memory efficient) is to store the fields normally, but make each field's lens clear out all the other fields when it's being set.

judah:
Resolves #21.

Also refactors (again) the logic around codegen for fields and oneofs, as a follow-up
to #85 which created that sum type in the first place.  (In part, it reverts
code structure to an earlier version of that PR.)

Suppose we have

    message Foo {
      oneof bar {
        int32 a = 1;
        int64 b = 2;
      }
    }

Then we'll generate data types:

    data Foo { _Foo'bar :: Maybe Foo'Bar }

    data Foo'Bar = Foo'A Int32 | Foo'B Int64

(Note: this changes from the previous behavior which called it Foo'bar; the
capitalization/camelcasing feels more appropriate since it's a datatype.)

and overloaded versions of the following lenses:

    maybe'bar :: Lens' Foo (Maybe Foo'Bar)
    a         :: Lens' Foo Int32
    maybe'a   :: Lens' Foo (Maybe Int32)
    b         :: Lens' Foo Int64
    maybe'b   :: Lens' Foo (Maybe Int64)

The `maybe'bar` accessor returns `Nothing` when the oneof isn't set, (or set
to an unknown case from a different version of the .proto file).  We don't
provide a straight `bar` accessor since there's no clear default between
different cases of the oneof.

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
@@ -31,7 +31,7 @@ module Data.ProtoLens.Compiler.Definitions
 
 import Data.Char (isUpper, toUpper)
 import Data.Int (Int32)
-import Data.List (mapAccumL, foldl')
+import Data.List (mapAccumL)
 import qualified Data.Map as Map
 import Data.Maybe (fromMaybe)
 import Data.Monoid
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
@@ -7,7 +7,9 @@
 -- | This module takes care of collecting all the definitions in a .proto file
 -- and assigning Haskell names to all of the defined things (messages, enums
 -- and field names).
-{-# LANGUAGE DeriveFunctor, OverloadedStrings #-}
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings #-}
 module Data.ProtoLens.Compiler.Definitions
     ( Env
     , Definition(..)
@@ -16,6 +18,9 @@ module Data.ProtoLens.Compiler.Definitions
     , OneofInfo(..)
     , OneofCase(..)
     , FieldName(..)
+    , Symbol
+    , nameFromSymbol
+    , promoteSymbol
     , EnumInfo(..)
     , EnumValueInfo(..)
     , qualifyEnv
@@ -31,7 +36,7 @@ import qualified Data.Map as Map
 import Data.Maybe (fromMaybe)
 import Data.Monoid
 import qualified Data.Set as Set
-import Data.String (fromString)
+import Data.String (IsString(..))
 import Data.Text (Text, cons, splitOn, toLower, uncons, unpack)
 import qualified Data.Text as T
 import Lens.Family2 ((^.))
@@ -58,7 +63,9 @@ import Data.ProtoLens.Compiler.Combinators
     ( Name
     , QName
     , ModuleName
+    , Type
     , qual
+    , tyPromotedString
     , unQual
     )
 
@@ -107,7 +114,7 @@ data OneofCase = OneofCase
     }
 
 data FieldName = FieldName
-    { overloadedName :: String
+    { overloadedName :: Symbol
       -- ^ The overloaded name of lenses that access this field.
       -- For example, if the field is called "foo_bar" in the .proto
       -- then @overloadedName == "fooBar"@ and we might generate
@@ -119,6 +126,27 @@ data FieldName = FieldName
       -- ^ The Haskell name of this internal record field; for example,
       -- "_Foo'Bar'baz.  Unique within each module.
     }
+
+-- | A string that refers to the name (in Haskell) of a lens that accesses a
+-- field.
+--
+-- For example, in the signature of the overloaded lens
+--
+-- @
+--     foo :: HasLens "foo" ... => Lens ...
+-- @
+--
+-- a 'Symbol' is used to construct both the type-level argument to
+-- @HasLens@ and the name of the function @foo@.
+newtype Symbol = Symbol String
+    deriving (Eq, Ord, IsString, Monoid)
+
+nameFromSymbol :: Symbol -> Name
+nameFromSymbol (Symbol s) = fromString s
+
+-- | Construct a promoted, type-level string.
+promoteSymbol :: Symbol -> Type
+promoteSymbol (Symbol s) = tyPromotedString s
 
 -- | All the information needed to define or use a proto enum type.
 data EnumInfo n = EnumInfo
@@ -200,7 +228,7 @@ messageDefs protoPrefix hsPrefix d
             }
     fieldInfo f = FieldInfo f $ mkFieldName $ f ^. name
     mkFieldName n = FieldName
-                    { overloadedName = n'
+                    { overloadedName = fromString n'
                     , haskellRecordFieldName = fromString $ "_" ++ hsPrefix' ++ n'
                     }
       where
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
@@ -39,7 +39,7 @@ import qualified Data.Set as Set
 import Data.String (IsString(..))
 import Data.Text (Text, cons, splitOn, toLower, uncons, unpack)
 import qualified Data.Text as T
-import Lens.Family2 ((^.))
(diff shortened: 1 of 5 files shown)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -78,32 +79,45 @@ data Definition n = Message (MessageInfo n) | Enum (EnumInfo n)
...
+    , caseConstructorName :: Name
+        -- ^ The constructor for building a 'oneofTypeName' from the
+        -- value in this field.
     }
 
-data OneofFieldInfo = OneofFieldInfo
-    { oneofConstructorName :: Name
-    , oneofEnclosingFieldName :: Name
+data FieldName = FieldName
+    { overloadedName :: String
```
isovector: why isn't this a `Name`?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -163,60 +177,67 @@ messageAndEnumDefs protoPrefix hsPrefix messages enums
...
-        in (oneofFields, OneofInfo
-               { oneofTypeName = typename
-               , oneofRecordFieldName = encFieldName
-               })
+
+-- | Group fields by the index of the oneof field that they belong to.
+-- (Or 'Nothing' if they don't belong to a oneof.)
+collectFieldsByOneofIndex
+    :: [FieldDescriptorProto] -> Map.Map (Maybe Int32) [FieldDescriptorProto]
+collectFieldsByOneofIndex =
```
isovector: how about

`collectFieldsByOneofIndex = Map.fromListWith (++) . fmap (view maybe'oneofIndex &&& pure)`

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -163,60 +177,67 @@ messageAndEnumDefs protoPrefix hsPrefix messages enums
...
+    allFields = collectFieldsByOneofIndex (d ^. field)
+    thisDef =
+        Message MessageInfo
+            { messageName = fromString $ hsPrefix ++ hsName (d ^. name)
+            , messageDescriptor = d
+            , messageFields =
+                  map fieldInfo $ Map.findWithDefault [] Nothing allFields
+            , messageOneofFields =
+                  map (uncurry oneofInfo)
+                      $ zip [0..] $ map (^. name) $ d ^. oneofDecl
```
isovector: `zipWith oneofInfo [0..]`?

isovector: not sure how you feel about lens tricks, but this could also be `d ^.. oneofDecl . traverse name`

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -205,7 +211,8 @@ generateMessageDecls syntaxType env protoName info =
     ]
   where
     dataType = tyCon $ unQual dataName
-    MessageInfo { messageName = dataName, messageFields = fields, messageOneofFields = oneofFields } = info
+    MessageInfo { messageName = dataName } = info
```
isovector: `dataName = messageName info`

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -369,103 +376,144 @@ generateFieldDecls xStr =
...
+        -- ^ A lens from the recordFieldType to the lensFieldType; i.e.,
+        -- from how it's actually stored in the Haskell record to how the
+        -- lens views it.
     }
 
 -- | Compile information about the record field type and type/class instances
 -- for this particular field.
-lensInfo :: SyntaxType -> Env QName -> FieldInfo -> LensInfo
-lensInfo syntaxType env f = case fd ^. label of
+plainRecordField :: SyntaxType -> Env QName -> FieldInfo -> RecordField
```
isovector: maybe update the haddock to explain what "plain" means

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -579,29 +627,29 @@ rawFieldAccessor f = "Lens.Family2.Unchecked.lens" @@ getter @@ setter
...
-                    (pVar "_otherwise")
-                    (con "Prelude.Nothing")
-                ]
-        setter = lambda ["x__", "y__"]
-                    $ recUpdate "x__" [fieldUpdate (unQual encName)
-                    $ ("Prelude.fmap" @@ (con $ unQual consName) @@ "y__")]
+--   (\ _ y__ -> fmap Foo'c y__
+oneofFieldAccessor :: OneofCase -> Exp
+oneofFieldAccessor
+    OneofCase { caseConstructorName = consName }
```
isovector: I'd prefer a view pattern here; record punning makes my head spin

`oneofFieldAccessor (caseConstructorName -> consName)`

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -579,29 +627,29 @@ rawFieldAccessor f = "Lens.Family2.Unchecked.lens" @@ getter @@ setter
...
-                    (pApp "Prelude.Just" [pApp (unQual consName) [pVar "x__val"]])
-                    ("Prelude.Just" @@ "x__val")
-                , alt
-                    (pVar "_otherwise")
-                    (con "Prelude.Nothing")
-                ]
-        setter = lambda ["x__", "y__"]
-                    $ recUpdate "x__" [fieldUpdate (unQual encName)
-                    $ ("Prelude.fmap" @@ (con $ unQual consName) @@ "y__")]
+--   (\ _ y__ -> fmap Foo'c y__
```
isovector: I don't understand this change -- can you elaborate please?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
judah: At some points (e.g., when we promote it to a type-level string) we actually need a raw `String`.  (Also, for some lenses we want to prefix it with `"maybe'"`.  The `Name` datatype is awkward to use for those cases.

Added a "Symbol" newtype to make the meaning more explicit.  (It should also help prevent some bugs that I ran into when developing this PR.)

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
judah: Done with `fromListWith`.  I decided to avoid `&&&/pure` since I feel the explicit lambda is a little more readable.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
judah: D'oh, missed the obvious `zipWith`.  (Interestingly, hint actually recommended it, I just missed its message.)

Thanks for explaining how to use `^..`; I tried using `traverse` earlier but couldn't figure out the right invocation.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
judah: Done.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
judah: Done.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
judah: Added more to the comment.

The main change is that previously when we had

    data Foo = Foo Foo'Bar
    data Foo'Bar = Case1 Int32 | Case2 Int64

this function generated a `Lens' Foo (Maybe Int64)`, i.e., both accessing the sum type and also accessing its case.  Now it just accesses the case of the sum type (`Lens' (Maybe Foo'Bar)` (Maybe Int64)`) and delegates the former to more generic code.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
judah: Simplified as above to a plain function call.

PR #93, opened 2017-07-23
Export the `arbitraryMessage` function.

judah:
Bumps the version to proto-lens-arbitrary-0.1.1.0.

It's fairly useful on its own for anywhere you want a
`Gen MyProtoMessage`; for example, when defining your own
custom `Arbitrary` instances for types that contain protobufs.

Also use `arbitraryMessage/shrinkMessage` a little more directly
in the internal implementation.


judah:
@isovector FYI

PR #95, opened 2017-07-25
Generate Ord instances for all exported datatypes

isovector:
This PR adds `deriving (Ord)` to all generated data types (messages, oneofs, enums).

The diff is bigger than I'd expect; presumably due to some previous change not having its bootstrap run.

### DIFF
--- a/proto-lens-tests/tests/enum.proto
+++ b/proto-lens-tests/tests/enum.proto
@@ -20,6 +20,13 @@ enum Baz {
   BAZ8 = 12;
 }
 
+message TwoBazs {
+  // We don't have the direct capability to generate arbitrary instances of
+  // enums, only of messages containing those enums.
+  required Baz baz1 = 1;
+  required Baz baz2 = 2;
+}
+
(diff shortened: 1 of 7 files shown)


isovector:
I signed it!

blackgnezdo:
@isovector could you rebase?

judah:
Thanks for making this change!

Can you please add a unit test making sure the Ord instance is consistent with the Enum instance?  It would be similar to: https://github.com/google/proto-lens/blob/master/proto-lens-tests/tests/enum_test.hs#L100

proto-lens-tests/tests/enum.proto
```diff
@@ -20,6 +20,13 @@ enum Baz {
   BAZ8 = 12;
 }
 
+message TwoBazs {
+  // We don't have the direct capability to generate arbitrary instances of
```
judah: Ah, good point.  Can you please open a PR?  (In the meantime it seems relatively straightforward to work around.)

Issue #98, opened 2017-07-26
Inconsistent naming of coproduct oneof fields


isovector:
```
message AcmeObservation {
  oneof status {
    ActionWin win = 2;
    CompletedHurdleStatus completed_hurdle = 3;
    QualifyTransaction qualify_transaction = 4;
  }
}
```

results in the generated haskell:

```haskell
data AcmeObservation'Status = AcmeObservation'Win !ActionWin
                            | AcmeObservation'Completed_hurdle !CompletedHurdleStatus
                            | AcmeObservation'Qualify_transaction !QualifyTransaction
                            deriving (Prelude.Show, Prelude.Eq)
```

Notice `AcmeObservation'Completed_hurdle`, which should become `AcmeObservation'CompletedHurdle` according to the renaming of all other snake case identifiers.

PR #99, opened 2017-07-26
CamelCase messages and oneof fields/types.

judah:
Fixes some corner cases: previously underlines (snake_case) were preserved in
message types and oneof types and field constructors.  Instead, we should
camel case everything consistently.


isovector:
Fixes #98

PR #100, opened 2017-07-28
Separate types into their own module, apart from field lenses.

isovector:
~~Besides the cabal changes, this should be completely backwards compatible. ~~

there is no backwards compatibility anymore


judah:
Thanks again for making this big change!

Issue #61, opened 2016-12-15
How to parse delimited messages


LeanderK:
I got a socket connection that is used to stream data with delimited protobuf messages (see [this](https://developers.google.com/protocol-buffers/docs/techniques#streaming)). In the java api there is an convenient method called [parseDelimitedFrom](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Parser#parseDelimitedFrom-java.io.InputStream-). How can this be achieved using this library?
I am new to haskell, so i might have overlooked something. I am sorry if this is very obvious.

judah:
Currently there's no way to do this automatically.  I'm assuming your stream is exposed as a Haskell `System.IO.Handle`?  You could implement it manually with the following steps:
- Read a varint from the steam
- Read a `ByteString` of that length from the stream (e.g., with `Data.ByteString.hGet`)
- Call `Data.ProtoLens.decodeMessage` on the returned `ByteString`.

I don't know if there's any standalone code for reading a varint (there probably ought to be).  You could try copying proto-lens's internal routine, for example adapting its inner loop to read bytes from a `Handle` instead of calling attoparsec's `anyWord` parser:
https://github.com/google/proto-lens/blob/master/proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs#L38

LeanderK:
ah ok, so there is not some obvious function i somehow missed. I hoped this would be an easy project to do some practical haskell stuff ;) It still looks doable, i just might try solving it. If i get it to work and am confident in the code, would an PR be welcome, or is this too much of an edge case?

judah:
Hm...too be honest, such a function might be better as a separate package, at least until proto-lens has a better story around streaming in general.  For example, we could provide

    parseMessage :: Message a => Parser a
    parseDelimitedMessage :: Message a => Parser a

and you could use `parseDelimitedMessage` along with a library like `attoparsec-conduit` to process a stream of inputs.  But it's not clear that we'll stick with `attoparsec` in the long term (#62), so I'm hesitant to expose such an API from the core of `proto-lens` at this time.

LeanderK:
ok, that makes sense! thank you.

judah:
This was done by #268.

LeanderK:
well, this was a long time ago! Thanks!

ulysses4ever:
[...]

PR #102, opened 2017-07-29
Add buildMessageDelimited: size-delimited streams of Messages

shlevy:


shlevy:
Note: This currently typechecks but is not tested, testing locally next.

judah:
Note that #61 has some prior discussion of this feature.

shlevy:
Tested to work via https://github.com/LukeHoersten/prometheus/pull/7

shlevy:
I signed it

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
```diff
@@ -57,11 +58,27 @@ anyBits = loop 0 0
...
+  mappend = (<>)
+
+putVarInt :: Word64 -> SizedBuilder
+putVarInt x = SizedBuilder sz $ putVarInt' x
+  where
+    sz = fromIntegral $ 1 + ((64 - countLeadingZeros x) `quot` 7)
+    putVarInt' n
+        | n < 128 = Builder.word8 (fromIntegral n)
+        | otherwise = Builder.word8 (fromIntegral $ n .&. 127 .|. 128)
+                      <> putVarInt' (n `shiftR` 7)
```
meiersi-da: @judah @shlevy note that you could use http://hackage.haskell.org/package/bytestring-0.10.8.2/docs/Data-ByteString-Builder-Prim.html to avoid the recursion completely and perform the check whether the buffer is large enough in a single step. See the `encodeCharUtf8HtmlEscaped` example for inspiration.

meiersi-da: Recalling the times when I was working on the bytestring builder, I'd expect about an order-of-magnitude speedup for encoding `varInt`s with this change.

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
```diff
@@ -57,11 +58,27 @@ anyBits = loop 0 0
             b <- anyWord8
             loop (w .|. shiftL (fromIntegral b) n) (n+8)
 
-putVarInt :: Word64 -> Builder
-putVarInt n
-    | n < 128 = Builder.word8 (fromIntegral n)
-    | otherwise = Builder.word8 (fromIntegral $ n .&. 127 .|. 128)
-                      <> putVarInt (n `shiftR` 7)
+-- | A wrapper around a 'Builder' that keeps track of its size.
+data SizedBuilder = SizedBuilder
```
meiersi-da: I would wary of the lazy `_size` as well. If we want to maximize encoding speed, I would expect that the approach used by http://hackage.haskell.org/package/cborg with an intermediate custom tailored token stream and a hot encoder loop would perform best. See: https://youtu.be/1WrSPSYhE6o?t=28m23s

shlevy:
@judah removed premature optimization

PR #104, opened 2017-07-30
Bump dependency to haskell-src-exts-0.19.

judah:
It will most likely be in the next stack LTS.

Tested by manually editing `stack.yaml` to depend on that version.


blackgnezdo:
Maybe update the summary to read haskell-src-exts?

blackgnezdo:
LGTM

PR #108, opened 2017-07-31
Prepare for releasing proto-lens-0.2.2.0.

judah:
- Update the changelogs (including updated changes to `proto-lens-protoc`).
- Make `proto-lens`, `proto-lens-descriptors` and `proto-lens-protobuf-types`
  depend on `0.2.2.*` rather than `0.2.2.0`.  That should make it easier
  to make bugfix package releases without updating all three packages in
  lockstep.

### DIFF
--- a/proto-lens-descriptors/proto-lens-descriptors.cabal
+++ b/proto-lens-descriptors/proto-lens-descriptors.cabal
@@ -27,7 +27,7 @@ library
     , data-default-class >= 0.0 && < 0.2
     , lens-family == 1.2.*
     , lens-labels == 0.1.*
-    -- Specify an exact version of `proto-lens`, since it's tied closely
-    -- to the generated code.
-    , proto-lens == 0.2.2.0
+    -- Specify a more precise version of `proto-lens`, since it depends on
+    -- the exact definition of the Message class.
+    , proto-lens == 0.2.2.*
     , text == 1.2.*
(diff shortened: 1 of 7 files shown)


proto-lens-descriptors/proto-lens-descriptors.cabal
```diff
@@ -27,7 +27,7 @@ library
     , data-default-class >= 0.0 && < 0.2
     , lens-family == 1.2.*
     , lens-labels == 0.1.*
-    -- Specify an exact version of `proto-lens`, since it's tied closely
-    -- to the generated code.
-    , proto-lens == 0.2.2.0
+    -- Specify a more precise version of `proto-lens`, since it depends on
+    -- the exact definition of the Message class.
```
blackgnezdo: Does it begin to feel like consolidating into fewer package would be beneficial? Configuration management is pretty dull...

judah: It really is a pain...I'm not sure how to solve bootstrapping in that case, though.  Filed #109.

judah: For what it's worth, I do still prefer separate packages for `proto-lens-combinators/arbitrary/optparse/etc`.  It's annoying when there's a new major version of GHC, but otherwise they've been updated pretty independently.

PR #114, opened 2017-08-02
Partial matching comparison

isovector:
This PR adds a `partial` function which returns true iff one message is partially the same as another (meaning regular equality for subfields, except for unset optional fields).

Haven't written any tests, since this scratches an itch I have in my day job and I'm keen to get back to it :)

It seems like it would make sense to pack the `Ord` and `Show` instances into the `FieldAccessor` GADT as well, giving anyone doing reflection the full gamut of available typeclasses generated by protolens.


isovector:
This is crap. I'll send a better one.

PR #118, opened 2017-08-06
Improve readability of HasLens instances.

judah:
- Swaps the order of type arguments to `HasLens` in order
  to make the instances more readable in Haddock and error
  messages.
- Adds a new class `HasLens'` which is effectively the
  "monomorphic" version of `HasLens`; that is, for lenses of
  monomorphic types (which includes all proto messages).

For example:

    instance (HasLens' f Foo x a, a ~ b) => HasLens f Foo Foo x a b where
        where lensOf = lensOf'
    instance Functor f => HasLens' f Foo "a" Int where ...
    instance Functor f => HasLens' f Foo "b" Double where ...
    instance Functor f => HasLens' f Foo "c" [Float]  where ...
    ...

And for a polymorphic type:

    instance Functor f => HasLens f (a,b) (a,c) "snd" b c where
        lensOf _ = lens snd (\(x,_) y -> (x,y))

### DIFF
--- a/lens-labels/src/Lens/Labels.hs
+++ b/lens-labels/src/Lens/Labels.hs
@@ -37,6 +37,7 @@ module Lens.Labels (
     HasLens(..),
     Proxy#,
     proxy#,
+    HasLens'(..),
     -- * Setters
     ASetter,
     (.~),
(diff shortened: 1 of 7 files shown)


judah:
Example improvement in error messages: https://gist.github.com/judah/e6678257cc28f2c486a631e49db885f2

Issue #27, opened 2016-07-13
Support extensions


judah:
Extensions (proto2-only) currently aren't supported yet.  It's not clear what the API should look like.

This would primarily be useful for legacy code, since proto3 replaces extensions with the `Any` type (#22).

gleber:
Judah, have you given any thoughts about this feature?

judah:
I haven't really had a chance to consider it in detail.  Here's a potential approach which is along the same lines as other language bindings:

First, define an analogue `RawMessage` of `google::protobuf::Message`, which essentially holds the raw tagged values in a `Map`.

Now, suppose we have the following code in `foo.proto`:

    // File foo.proto
    message Foo {
        extensions 100 to 199;
    }

We add a new field to `Foo` with name `_Foo__extensions` and type `RawMessage`.  (This approach would also essentially solve #29.)

Finally, suppose we have in `bar.proto`:

    extend Foo {
       optional int32 bar = 123;
    }

    message Baz {
        extend Foo {
            optional Baz foo_ext = 124;
        }
    }

Then we define two new instances of `Lens.Label.HasLens`: `HasLens "bar" f Foo Foo Int32 Int32`, and `HasLens "foo_ext" f Foo Foo Baz Baz`; where the implementation of each lens uses the `RawMessage` that `Foo` holds.

PR #124, opened 2017-08-23
Add support for parsing `Any` messages in google protobuf text format

tathougies:
This PR implements a special `Registry` object (thanks @judah) that contains a mapping between type URIs and MessageDescriptors. We can create instances of these data types using the `register` function and combine them monadically. The PR implements `showMessageWithRegistry` and `readMessageWithRegistry` which allows the text format parser to read text-formatted messages containing `Any` fields. The `readMessageWithRegistry` function will read and pack such fields into the standard `Any` object. If the type URI can't be found in the registry, an error is emitted. The `showMessageWithRegistry` function uses the registry to encode messages with embedded `Any`s using the more human-readable `Any` syntax. If it encounters an `Any` whose type URI cannot be found in the registry, the `Any` is printed in the packed format.

### DIFF
--- a/proto-lens-tests/proto-lens-tests.cabal
+++ b/proto-lens-tests/proto-lens-tests.cabal
@@ -32,6 +32,8 @@ library
                -- compile tests/package-deps.proto (since it's listed in
                -- extra-source-files regardless).
                , proto-lens-tests-dep
+               -- For tests/imports.proto:
+               , proto-lens-descriptors
                , base >= 4.8 && < 4.11
                , bytestring == 0.10.*
                , text == 1.2.*
--- a/proto-lens-tests/proto-lens-tests.cabal
+++ b/proto-lens-tests/proto-lens-tests.cabal
@@ -162,6 +162,7 @@ Test-Suite text_format_test
                , proto-lens
                , proto-lens-protoc
                , proto-lens-tests
+               , proto-lens-protobuf-types
                , test-framework
                , test-framework-hunit
                , text
--- a/proto-lens-tests/proto-lens-tests.cabal
+++ b/proto-lens-tests/proto-lens-tests.cabal
@@ -32,6 +32,7 @@ library
                -- compile tests/package-deps.proto (since it's listed in
                -- extra-source-files regardless).
                , proto-lens-tests-dep
+               , proto-lens-protobuf-types
                -- For tests/imports.proto:
                , proto-lens-descriptors
                , base >= 4.8 && < 4.11
@@ -161,8 +162,8 @@ Test-Suite text_format_test
                , pretty
                , proto-lens
                , proto-lens-protoc
-               , proto-lens-tests
                , proto-lens-protobuf-types
+               , proto-lens-tests
                , test-framework
                , test-framework-hunit
                , text
--- a/proto-lens-tests/tests/text_format.proto
+++ b/proto-lens-tests/tests/text_format.proto
@@ -2,6 +2,8 @@ syntax = "proto2";
 
 package text_format;
 
+import "google/protobuf/any.proto";
+
 message Test1 {
   // Note: the field tag numbers are intentionally ordered differently
   // than the names, to make sure we print ordered by tag number.
@@ -13,4 +15,9 @@ message Test1 {
 
 message Test2 {
   optional Test1 c = 1;
+}
+
+message AnyHolder {
+  optional google.protobuf.Any thing1 = 1;
+  optional google.protobuf.Any thing2 = 2;
 }
--- a/proto-lens-tests/tests/text_format.proto
+++ b/proto-lens-tests/tests/text_format.proto
@@ -19,5 +19,8 @@ message Test2 {
 
 message AnyHolder {
   optional google.protobuf.Any thing1 = 1;
-  optional google.protobuf.Any thing2 = 2;
+  message Sub {
+    optional google.protobuf.Any thing2 = 1;
+  }
+  optional Sub sub = 2;
 }
--- a/proto-lens/src/Data/ProtoLens/Message.hs
+++ b/proto-lens/src/Data/ProtoLens/Message.hs
@@ -10,6 +10,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE PatternGuards #-}
 -- | Datatypes for reflection of protocol buffer messages.
 module Data.ProtoLens.Message (
     -- * Reflection of Messages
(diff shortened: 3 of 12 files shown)


tathougies:
I signed it!

proto-lens/src/Data/ProtoLens/Message.hs
```diff
@@ -231,3 +260,17 @@ reverseRepeatedFields fields x0
     reverseListField x (FieldDescriptor _ _ (RepeatedField _ f))
         = over f reverse x
     reverseListField x _ = x
+
+newtype Registry = Registry (Map.Map T.Text SomeMessageType)
```
judah: Let's add Haddocks for these functions, maybe:

    -- | A set of known message types.  Can help encode/decode protobufs containing
    -- @Data.ProtoLens.Any@ values in a more human-readable text format.
    -- 
    -- Registries can be combined using their 'Monoid' instance. 
    newtype Registry = ...

    -- | Build a 'Registry' containing a single proto type.
    register :: ...

    -- | Look up a message type by name (e.g., @"google.protobuf.FloatValue"@).
    -- The URL corresponds to the field @google.protobuf.Any.type_url@.
    lookupRegistered :: ...

proto-lens/src/Data/ProtoLens/TextFormat.hs
```diff
@@ -203,37 +236,55 @@ modifyField (RepeatedField _ f) value = over f (value :)
...
-makeValue EnumField (Parser.EnumValue x) =
+makeValue _ EnumField (Parser.EnumValue x) =
     maybe (Left $ "Unrecognized enum value " ++ show x) Right
         (readEnum x)
-makeValue MessageField (Parser.MessageValue x) = buildMessage x
-makeValue GroupField (Parser.MessageValue x) = buildMessage x
-makeValue f val = Left $ "Type mismatch parsing text format: " ++ show (f, val)
+makeValue reg field@MessageField val@(Parser.MessageValue typeUri x)
+    | Just AnyMessageDescriptor { anyTypeUrlLens, anyValueLens } <- matchAnyMessage field =
+        case typeUri of
```
judah: Can we move these pattern matches into to the guard?  If decoding the Any fails, we should just default to displaying the raw typeUrl and value (i.e. l.282) rather than failing the decode.

Also, maybe add a test for that edge case, if it's straightforward.

proto-lens-tests/tests/text_format.proto
```diff
@@ -14,3 +16,8 @@ message Test1 {
 message Test2 {
   optional Test1 c = 1;
 }
+
+message AnyHolder {
+  optional google.protobuf.Any thing1 = 1;
+  optional google.protobuf.Any thing2 = 2;
```
tathougies: good idea!

proto-lens/src/Data/ProtoLens/Message.hs
tathougies: Sure

judah:
I re-ran the Travis build and was able to get it running.  Unfortunately it failed for a couple of reasons:

- The use of TypeApplications (see my earlier message)
- The use of pattern guards (I guess it needs an explicit `LANGUAGE PatternGuards`):
https://travis-ci.org/google/proto-lens/jobs/267708755

Can you try fixing those and rebuilding with `stack build --pedantic`?

tathougies:
Great... will do!

proto-lens/src/Data/ProtoLens/TextFormat.hs
tathougies: Can you explain what you mean here? This is for decoding a message which has been explicitly given a URL. I don't think there's any other meaning here other than wanting to encode an Any?

judah: Ah, sorry, I got confused.  Here's a more coherent concern: if the message is in the registry but we see a raw Any (e.g. `typeUri: ...  value: "..."`) then it still ought to decode it successfully.

Something like:

    makeValue reg MessageField val@(Parser.MessageValue Nothing x)
        = buildMessage reg x
    makeValue reg field val@(Parser.MessageValue (Just uri) x)
        | Just AnyMessageDescriptor { anyTypeUrlLens, anyValueLens } <- matchAnyMessage field
            = case lookupRegistered uri reg of
                  Nothing -> Left $ "Could not decode Any of unknown type " ++ show uri
                  Just (SomeMessageType (Proxy :: Proxy value')) -> ...

proto-lens/src/Data/ProtoLens/TextFormat/Parser.hs
```diff
@@ -67,7 +69,7 @@ data Key = Key String  -- ^ A standard key that is just a string.
 data Value = IntValue Integer  -- ^ An integer
   | DoubleValue Double  -- ^ Any floating point number
   | ByteStringValue ByteString    -- ^ A string or bytes literal
-  | MessageValue Message  -- ^ A sub message
+  | MessageValue (Maybe StrictText.Text) Message  -- ^ A sub message
```
judah: Comment on the constructor what its meaning is.

tathougies:
Any idea on why the CI keeps failing?

proto-lens-tests/proto-lens-tests.cabal
```diff
@@ -162,6 +162,7 @@ Test-Suite text_format_test
                , proto-lens
                , proto-lens-protoc
                , proto-lens-tests
+               , proto-lens-protobuf-types
```
judah: You'll need to add this to the dependencies of the *library*, not the test.  Sorry for this confusing behavior; see also the comment at l.29.

proto-lens/src/Data/ProtoLens/TextFormat.hs
```diff
@@ -203,37 +237,52 @@ modifyField (RepeatedField _ f) value = over f (value :)
...
+makeValue reg field@MessageField (Parser.MessageValue (Just typeUri) x)
+    | Just AnyMessageDescriptor { anyTypeUrlLens, anyValueLens } <- matchAnyMessage field =
+        case lookupRegistered typeUri reg of
+          Nothing -> Left "Could not decode Any"
+          Just (SomeMessageType (Proxy :: Proxy value')) ->
+            case buildMessage reg x :: Either String value' of
+              Left err -> Left err
+              Right value' -> Right (def & anyTypeUrlLens .~ typeUri
+                                         & anyValueLens .~ encodeMessage value')
+    | typeUri == messageName (descriptor :: MessageDescriptor value) =
```
judah: Remove this case.  The syntax of `[pkg.Foo]{...}` (i.e.without the `type.googleapis.com/` part) refers to proto2 "extensions" which are not yet implemented in proto-lens (#27):
https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L402

And in general the type of a proto message is different than the type of an Any that wraps it, so we should avoid confusing the two.

proto-lens/src/Data/ProtoLens/TextFormat.hs
```diff
@@ -52,28 +59,38 @@ import qualified Data.ProtoLens.TextFormat.Parser as Parser
 
 -- | Pretty-print the given message into a human-readable form.
 pprintMessage :: Message msg => msg -> Doc
-pprintMessage = pprintMessage' descriptor
+pprintMessage = pprintMessageWithRegistry mempty
+
+-- | Pretty-print the given message into human-readable form, using the given
+-- 'Registry' to decode @google.protobuf.Any@ values.
```
judah: s/decode/encode

tathougies: Looking back at the old reviews, you told me to change encode to decode and decode to encode :) Can we just leave it? 

In reality, it's doing both. I could say 'to encode Any values into the text format'

judah: Sure, sorry about that.  Whichever one makes more sense to you.

PR #126, opened 2017-08-25
Don't generate Haskell modules if they won't be used.

judah:
Previously we generated Haskell modules for every `*.proto` file listed
in `extra-source-files`, even if the resulting `Proto.*` module wasn't
referenced in the `.cabal` file.  This led to a few edge cases which
had confusing behavior, for example:

- If a `Proto.*` module was only used in a test but not the library,
  and depended on a `.proto` in another package.
- If a module was accidentally omitted from the `exposed-modules` list

Now we only generate modules that we know will actually be used
by some component in the `.cabal` file.


proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -296,6 +334,27 @@ findExecutableOrDie name debugMsg = do
             hPutStrLn stderr sep
             error $ "Missing executable " ++ show name
 
+-- | Collect all the module names that we need to build.
+-- For example: only include test-suites if we're building with tests enabled
+-- (e.g., `stack test` vs `stack build`).
+collectActiveModules :: LocalBuildInfo -> [ModuleName]
+collectActiveModules l = let
+    in (activeLib >>= exposedModules)
+        ++ concatMap otherModules
```
isovector: `mconcat`?

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -142,36 +179,37 @@ generatingProtos root = generatingSpecificProtos root getProtos
...
+    -> (LocalBuildInfo -> IO [FilePath])
     -- ^ A function to return a list of .proto files. Takes the Cabal package
     -- description as input. Non-absolute paths are treated as relative to the
     -- provided root directory.
     -> UserHooks -> UserHooks
 generatingSpecificProtos root getProtos hooks = hooks
-    { buildHook = \p l h f -> generate p l >> buildHook hooks p l h f
-    , haddockHook = \p l h f -> generate p l >> haddockHook hooks p l h f
-    , replHook = \p l h f args -> generate p l >> replHook hooks p l h f args
+    { buildHook = \p l h f -> generate l >> buildHook hooks p l h f
```
isovector: this is yucky, but i don't have a helpful suggestion

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
judah: Done.

PR #131, opened 2017-09-01
Update README now that we preserve unknown fields.

judah:


README.md
```diff
@@ -98,8 +98,6 @@ will generate the haskell files `Proto/Project/{Foo,Bar}.hs`.
 - Extensions (proto2-only) are not supported.  `Any` messages (the proto3
   equivalent) can be used, but don't have any custom API support like in the C++
   libraries.
-- Unknown fields of proto2 messages are discarded during decoding. (This is the
-  correct behavior for proto3.)
```
isovector: is this still the behavior for proto3 protolens?

judah: Added a link with more context.  They're changing the behavior of proto3 to support unknown fields.

Also updated a few more of those items since the list was pretty out of date.

PR #134, opened 2017-09-07
Partial matcher for protos

isovector:
This PR adds a function `partial` which performs a partial match between two protos.

### DIFF
--- a/proto-lens-tests/tests/partial_test.hs
+++ b/proto-lens-tests/tests/partial_test.hs
@@ -10,7 +10,7 @@ module Main where
 import qualified Data.Map as M
 import Data.ProtoLens
 import Data.ProtoLens.Arbitrary
-import Data.ProtoLens.Message (partial)
+import Data.ProtoLens.Compare (partial)
 import Lens.Family2 ((&), (.~))
 import Test.Framework (testGroup)
 import Test.Framework.Providers.HUnit (testCase)
--- a/proto-lens-tests/tests/partial_test.hs
+++ b/proto-lens-tests/tests/partial_test.hs
@@ -55,12 +55,12 @@ testMatchRepeated = makeTestCase "repeated"
                        , def & name .~ "3"
                              & another .~ 3
                        ])
-    (def @Foo & rep .~ [ def & name .~ "1"
-                             & another .~ 1
+    (def @Foo & rep .~ [ def & name .~ "3"
+                             & another .~ 3
                        , def & name .~ "2"
                              & another .~ 2
-                       , def & name .~ "3"
-                             & another .~ 3
+                       , def & name .~ "1"
+                             & another .~ 1
                        ])
 
 testMatchMap = makeTestCase "map"
--- a/proto-lens-tests/tests/partial_test.hs
+++ b/proto-lens-tests/tests/partial_test.hs
@@ -1,0 +1,98 @@
+-- Copyright 2016 Google Inc. All Rights Reserved.
+--
+-- Use of this source code is governed by a BSD-style
+-- license that can be found in the LICENSE file or at
+-- https://developers.google.com/open-source/licenses/bsd
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+module Main where
+
+import qualified Data.Map as M
+import Data.ProtoLens
+import Data.ProtoLens.Arbitrary
+import Data.ProtoLens.Message (partial)
+import Lens.Family2 ((&), (.~))
+import Test.Framework (testGroup)
+import Test.Framework.Providers.HUnit (testCase)
+import Test.HUnit.Base (assertBool)
+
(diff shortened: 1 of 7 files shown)


proto-lens-tests/tests/partial_test.hs
```diff
@@ -0,0 +1,98 @@
...
+    , testAlwaysMatchSame
+    , testMatchRequired
+    , testMatchOptional
+    , testMatchRepeated
+    ]
+
+makeTestCase :: Message a => String -> a -> a -> Test
+makeTestCase testName against match = testCase testName $ do
+    assertBool "matched"         $ partial against match
+    assertBool "not const" . not $ partial match against
```
judah: What does "not const" mean?

proto-lens-tests/tests/partial_test.hs
```diff
@@ -0,0 +1,98 @@
...
+testMatchRequired = makeTestCase "required"
+    (def @FooReq & bar . deep . name .~ "deeper")
+    (def @FooReq & ok .~ 5
+                 & bar . name .~ "hello"
+                 & bar . deep . name .~ "deeper")
+
+testMatchOptional = makeTestCase "optional"
+    (def @Foo & ok .~ 5)
+    (def @Foo & ok .~ 5
+                         & bar . name .~ "hello"
```
judah: Nit: align `&`s with the previous line.

proto-lens/src/Data/ProtoLens/Compare.hs
```diff
@@ -0,0 +1,108 @@
...
+    ) where
+
+import Data.Constraint (Dict (..))
+import qualified Data.Map as Map
+import Lens.Family2 (Lens', over, set, view)
+import Lens.Family2.Unchecked (lens)
+import Data.ProtoLens.Message
+
+-- | Perform a partial match on a protobuffer. This corresponds to equality for
+-- primitive fields, (recursive) partial equality for required 'Message'
```
judah: We should probably be more explicit in the comment: this is equality for proto3 scalar optional fields and proto2 required fields.

proto-lens/src/Data/ProtoLens/Compare.hs
```diff
@@ -0,0 +1,108 @@
...
+-- fields, partial equality if both are set for optional fields, and partial
+-- equality of subsets for repeated and map fields.
```
judah: I have mixed feelings about subset equality for repeated fields and whether that satisfies the principle of least surprise.  I can recall a couple tests I've written where I did want to specify the explicit size/order of the list, I just didn't care about every field of those values.

proto-lens/src/Data/ProtoLens/Compare.hs
```diff
@@ -0,0 +1,108 @@
...
+
+        RepeatedField _ l ->
+          and . fmap (\a -> any (eq ftd a) $ view l match) $ view l against
+
+        MapField k v l    ->
+          let entriesMatch   = toRepeatedEntry k v $ view l match
+              entriesAgainst = toRepeatedEntry k v $ view l against
+           in and $ fmap (\a -> any (eq ftd a) entriesMatch) entriesAgainst
+  where
+    toRepeatedEntry :: (Ord key, Message entry)
```
judah: Break these helper functions out into separate, top-level (unexported) functions.

Ok if it makes sense to group them together; for example, make `getEq` part of the where clause of `eq`.

proto-lens-tests/tests/partial_test.hs
isovector: it means `partial` isn't implemented as `\_ _ -> True`  :)

proto-lens/src/Data/ProtoLens/Compare.hs
isovector: yeah, iirc you somehow compose matchers at google to decide which of these behaviors you want

proto-lens/src/Data/ProtoLens/Compare.hs
```diff
@@ -0,0 +1,108 @@
...
+import qualified Data.Map as Map
+import Lens.Family2 (Lens', over, set, view)
+import Lens.Family2.Unchecked (lens)
+import Data.ProtoLens.Message
+
+-- | Perform a partial match on a protobuffer. This corresponds to equality for
+-- primitive fields, (recursive) partial equality for required 'Message'
+-- fields, partial equality if both are set for optional fields, and partial
+-- equality of subsets for repeated and map fields.
+partial
```
judah: One other request: can we call this something more specific like `partialEq`?  The word `partial` has an overloaded meaning (e.g. a function like `head`)...

judah:
Looks like this failed in CI for a couple reasons:
1. Warnings; please re-run with `stack test --pedantic`
2. `TypeApplications` isn't supported in 7.10.  It might be enough to just wait for #136.

judah:
Closing since this PR has become pretty stale.

PR #135, opened 2017-09-07
Remove obsolete files from `proto-lens-descriptors.

judah:
They were committed accidentally as part of #100.  (The `*'Fields` modules are
used instead.)


judah:
@isovector PTAL

isovector:
sorry!! lgtm

PR #136, opened 2017-09-07
Bundle enum pattern synonyms exports with their type.

judah:
This change removes support for `ghc-7.10`.  The feature we're using
was only added in `ghc-8.0.1`, and there are already a few other shims for
that older version of GHC.

This change uses explicit export lists in the generated proto modules (but
only for `Proto.Foo.Bar`, not `Proto.Foo.Bar'Fields`).  The same names should
be exported as before, the only difference is that for enum aliases:
```
enum Foo {
  option allow_alias = true;
  Foo = 1;
  Bar = 2;
  BarA = 2;
}
```
which produce the Haskell code:
```
data Foo = Foo | Bar
pattern BarA :: Foo
pattern BarA = Bar
```
we now export explicitly `Foo(.., BarA)`, rather than exporting `BarA` as a
separate name.  That way, if someone writes `import Proto.File (Foo(..))` then
they also get `BarA` automatically.


judah:
Good call, done.  I've bumped all the lower bounds to `base-4.9`.

judah:
I left the README alone since (a) the version bounds in the cabal files will give a pretty clear indication of what versions we're compatible with, and (b) so that's not yet another place we have to keep up to date with everything else.  But let me know if you feel strongly about it.

blackgnezdo:
I imagine you eventually may want to put a note into the changelog. LGTM otherwise.

PR #137, opened 2017-09-07
Proto3 enums

judah:
Implement proto3-style "open" enums.

For example: 
``` enum Foo {   Foo0 = 0;   Foo1 = 1; } ``` 
becomes
``` newtype Foo = Foo Int32
pattern Foo0 :: Foo
pattern Foo0 = Foo 0
pattern Foo1 :: Foo
pattern Foo1 = Foo 1 ```
Additionally, proto3 enums are exported from the `Proto.*` module with the form `Foo(Foo0, Foo1)`, so that they can be imported as `Foo(..)` to get the patterns.

This PR depends on #136.

### CODE REVIEW

**Comment by isovector** [comment]
i think you want the description to be

```haskell
pattern Foo0 :: Foo
pattern Foo0 = Foo 0
```

proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
```diff
@@ -161,16 +174,21 @@ match n ps e = Syntax.Match () n ps (Syntax.UnGuardedRhs () e) Nothing
...
+module' modName exports
     = Syntax.Module ()
         (Just $ Syntax.ModuleHead () modName
                     -- no warning text
                     Nothing
-                    -- no explicit exports; we export everything.
-                    -- TODO: Also export public imports, taking care not to
-                    -- cause a name conflict between field accessors.
-                    Nothing)
+                    (Syntax.ExportSpecList () <$> exports))
```
isovector: what's the reason for this change?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -236,11 +248,94 @@ generateMessageDecls syntaxType env protoName info =
...
+
+    -- proto3 enums always default to zero.
+    , instDecl [] ("Data.Default.Class.Default" `ihApp` [dataType])
+        [[match "def" [] $ "Prelude.toEnum" @@ litInt 0]]
+    , instDecl [] ("Data.ProtoLens.FieldDefault" `ihApp` [dataType])
+        [[match "fieldDefault" [] $ "Prelude.toEnum" @@ litInt 0]]
+    ]
+    ++
+    -- pattern Value0 :: Foo
+    -- pattern Value0 = Foo 0
```
isovector: <3

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -236,11 +248,94 @@ generateMessageDecls syntaxType env protoName info =
...
+          ]
+        | v <- enumValues info
+        , let n = enumValueName v
+        , let k = fromIntegral $ enumValueDescriptor v ^. number
+        ]
+  where
+    dataName = enumName info
+    dataType = tyCon $ unQual dataName
+
+generateEnumDecls Proto2 info =
```
isovector: what happens to dependent haskell code if you change the proto version? would anyone ever do that? if so, do these separate representations cause any problems?

proto-lens-tests/tests/proto3.proto
```diff
@@ -17,8 +17,10 @@ message Foo {
   Sub sub = 5;
 
   enum FooEnum {
+    option allow_alias = true;
```
isovector: :+1:

proto-lens-tests/tests/proto3_test.hs
```diff
@@ -98,6 +98,26 @@ main = testMain
       , testCase "string" $ (def :: Strings) @=? (def & string .~ "")
       , testCase "enum" $ (def :: Foo) @=? (def & enum .~ Foo'Enum1)
       ]
+  -- Enums are all pattern aliases
+  , testGroup "enum"
+      [ testCase "aliases are exported" $ Foo'Enum2 @=? Foo'Enum2a
+      , testCase "enum values" $ do
+          map toEnum [0, 3, 3] @=? [Foo'Enum1, Foo'Enum2, Foo'Enum2a]
+          ["Foo'Enum1", "Foo'Enum2", "Foo'Enum2", "toEnum 5"]
+              @=? map show [Foo'Enum1, Foo'Enum2, Foo'Enum2a, toEnum 5]
```
isovector: is `toEnum` the only way to construct lax enums?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
judah: It was part of #136 (now merged).  Sorry for the noise.

Previously the modules that we generated always exported everything via `module M where`, so the export list parameter was always `Nothing`.  Now, for the types module we need to pass an explicit list, i.e., `module M (...) where` instead, so it can export pattern synonyms in a nicer way.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
judah: If you change from proto2 to proto3 you could incomplete patterns, e.g. in
```
case MyEnum of
   Foo -> True
   Bar -> False
```
when the enum becomes an open type.  But `-Wall` or similar will warn about that case.  And given that this is an explicit behavior difference between proto2/proto3, I think it's something you'd want to be careful about in any language (similar to, e.g., the handling of default values).

proto-lens-tests/tests/proto3_test.hs
judah: We're also exporting the newtype constructor itself.

PR #138, opened 2017-09-10
Switch to hpack.

judah:
`stack build` will generate `*.cabal` files automatically from
`package.yaml`.  The generated files are now `gitignore`'d.

For reference: https://github.com/sol/hpack

Advantages:
- Autogenerates a bunch of `.cabal` file fields, which helps given the number of packages we're managing.
- Allows specifying package dependencies (and version numbers) only once in the file, rather than separately for each target.  In particular, `proto-lens-tests` gets a *lot* simpler.
- More consistent file format (e.g. with respect to sub-lists)

Using `hpack` doesn't add any new dependencies since `stack` already has support built-in as a library.  The generated `.cabal` file will be automatically included in source distributions
when uploaded to Hackage.

This change was mostly painless, if tedious.  I used `hpack-convert`
and then cleaned up the files for formatting, dependency organization,
pointing to github and protoc codegen idiosyncracies.

The only real changes were:
- Move `protoc-gen-haskell.hsc` into its own directory, and
  simplify the Cabal structure (and bootstrapping procedure)
  around the library and binary.
- Slightly reorganize the data-files in
  proto-lens-descriptors, since `hpack` doesn't support a `data-dir`
  parameter.


blackgnezdo:
I'm in general supportive of this change as I was never a big fan of cabal format.

Could you summarize the features that you found most persuasive in the PR description? There's a level of indirection and an extra dependency that should be paying for itself with some interest.

One concern is long term maintenance of hpack and its tracking of new cabal features. I guess we can always convert back and hopefully doing it just for a subset of packages is a supported scenario.

judah:
I've updated the description, PTAL.

We can definitely use plain `.cabal` files for a subset of the packages, if they run into issues.  (Though for consistency I think it's for everything use `hpack` when possible.)

PR #139, opened 2017-09-17
Split the `Message` class into separate methods.

judah:
Removes the `MessageDescriptor` datatype which was getting a little unwieldy.
Turn its elements (e.g. `fieldsByTag`) into individual methods of the
`Message` class.  We can just use `Proxy msg` for the same purpose as
`MessageDescriptor`.


blackgnezdo:
LGTM

Issue #109, opened 2017-07-31
Consolidate packages


judah:
The `proto-lens`, `proto-lens-protoc` and `proto-lens-descriptors` packages are tied pretty closely together.  Consider consolidating some or all of them.

The main concern is bootstrapping.  Changing the internals of `lens-labels` or `proto-lens` effectively breaks `proto-lens-descriptors` until the descriptor modules can be regenerated -- but regenerating them requires a working `proto-lens-protoc`, which depends on `proto-lens-descriptors`, introducing a cycle.  The current bootstrap script solves this using the fact that they're all separate packages: it builds a new `proto-lens-protoc` against an old, working version of `lens-labels`, `proto-lens` and `proto-lens-descriptors`, and uses that compiler to generate the new descriptor modules.  I'm not sure how to implement that process if they're all in the same Cabal package.

PR #140, opened 2017-09-17
Consolidate proto-lens-descriptors into proto-lens.

judah:
There's no real reason for the two packages to be separate,
except that it increases the library size of `proto-lens`.  But since
`proto-lens-descriptors` is usually a transitive dependency anyway (through
`proto-lens-protoc`) I don't think it matters much.

Also remove an old bootstrapping-related flag that's not
used anymore.  (It should have been removed in #138.)

Helps address #109.

Ideally we'd merge `proto-lens-protoc` into the `proto-lens` package as well,
but I'm not sure yet how to do that without making bootstrapping more
complicated.


judah:
The next Hackage release will be a major version bump of `proto-lens` and `proto-lens-descriptors` (0.2.* -> 0.3) due to some backwards incompatibilities.  The currently-released `proto-lens-*` all have bounds `< 0.3`.  Hopefully, this is enough for Cabal to prevent a built plan containing an old `proto-lens-descriptors` and the new `proto-lens`.

We could also mark `proto-lens-descriptors` as deprecated once we make this release; I'm not exactly sure how, but I've seen that done for some other packages.

judah:
Also, it looks like `proto-lens-protoc` is the only package on Hackage that depends on `proto-lens-descriptors`:
https://packdeps.haskellers.com/reverse/proto-lens-descriptors

PR #147, opened 2017-09-29
Refactor `FieldTypeDescriptor`.

judah:
Distinguish the recursive sub-message case from the other
scalar field types.

Motivated by the changes in #134.

Small behaivor change: this prevents any message type from being
packed, not just groups.  (But that's closer to the standard anyway.)

### DIFF
--- a/proto-lens-arbitrary/Changelog.md
+++ b/proto-lens-arbitrary/Changelog.md
@@ -3,8 +3,7 @@
 ## v0.1.1.2
 - Remove support for `ghc-7.10`. (#136)
 - Use a `.cabal` file that's auto-generated from `hpack`. (#138)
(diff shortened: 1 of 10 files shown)


judah:
It's sort of a subtle improvement, but given that we've already made breaking changes to the `Message` class I'd rather get this one over with at the same time.

The reason: often, code that iterates over the message descriptor treats recursive submessages differently than scalar fields, and in fact treats all the scalar fields in a similar way.  Having the two cases "flattened" into the same type like they were before led to awkward patterns; see for example all the `_`s here which are passing a value that's irrelevant for scalar fields:
https://github.com/google/proto-lens/blob/31361c67be6e36519a9a586d5ca96f2ba6d2a9f7/proto-lens/src/Data/ProtoLens/TextFormat.hs#L119

Issue #150, opened 2017-10-20
Change proto3 enums (back) to a sum type


judah:
Description of proto3 enums (reformatted from [the docs](https://developers.google.com/protocol-buffers/docs/proto3)):

> During deserialization, unrecognized enum values will be preserved in the message, though how this is represented when the message is deserialized is language-dependent.
> 1. In languages that support open enum types with values outside the range of specified symbols, such as C++ and Go, the unknown enum value is simply stored as its underlying integer representation.
> 2. In languages with closed enum types such as Java, a case in the enum is used to represent an unrecognized value, and the underlying integer can be accessed with special accessors. In either case, if the message is serialized the unrecognized value will still be serialized with the message.

Currently (i.e., on HEAD) we're using option #1.  That is, if we had `enum Foo = { A = 1; B = 2 }` then we generate `newtype Foo = Foo Int32` and define `A` and `B` as pattern synonyms:
```
pattern A = Foo 1
pattern B = Foo 2
```
This is simpler, but limits our ability to get exhaustiveness checking from the compiler.  Specifically,
 if someone adds a new enum case to the proto, the type checker won't tell us that we're now missing a case.  This issue happened to us in real code.

GHC 8.2.1 does has `COMPLETE` directives for pattern synonyms, but (a) it's too soon to drop support for 8.0, and (b) that's a newer and less-well-understood feature.

------

The proposal for the new API is similar to what already exists for Scala and Java.  For:
```enum Foo = { A = 1; B = 2; }```
generate the following code:
```
data Foo = A | B | Foo'Unrecognized Foo'UnrecognizedValue

-- | Representation of an unknown value.  Uses a newtype to make
-- the different branches of `Foo` provably distinct.
-- For example, this way we don't have to worry about whether
-- `A == Foo'Unrecognized (Foo'UnrecognizedValue 1)`.
newtype Foo'UnrecognizedValue = Foo'UnrecognizedValue Int32 -- hidden constructor

unrecognizedValue'Foo :: Foo'UnrecognizedValue -> Int32

instance Enum Foo where
    toEnum 1 = A
    toEnum 2 = B
    toEnum n = Foo'Unrecognized (Foo'UnrecognizedValue n)
    fromEnum = ...
```

PR #151, opened 2017-10-26
Sumtype enums

pkinsky:
#150

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
@@ -358,8 +358,3 @@ string s = Syntax.String () s (show s)
 modifyModuleName :: (String -> String) -> ModuleName -> ModuleName
 modifyModuleName f (Syntax.ModuleName _ unpacked) =
   Syntax.ModuleName () $ f unpacked
-
-
-modifyIdent :: (String -> String) -> Name -> Name
-modifyIdent f (Syntax.Ident l s) = Syntax.Ident l $ f s
-modifyIdent _ x = x
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
@@ -166,6 +166,9 @@ ihApp = foldl (Syntax.IHApp ())
 tyParen :: Type -> Type
 tyParen = Syntax.TyParen ()
 
+paren :: Exp -> Exp
+paren = Syntax.Paren ()
+
 type Match = Syntax.Match ()
 
 -- | A simple clause of a function binding.
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
@@ -361,3 +361,4 @@ string s = Syntax.String () s (show s)
 modifyModuleName :: (String -> String) -> ModuleName -> ModuleName
 modifyModuleName f (Syntax.ModuleName _ unpacked) =
   Syntax.ModuleName () $ f unpacked
+
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
@@ -359,3 +359,7 @@ modifyModuleName :: (String -> String) -> ModuleName -> ModuleName
 modifyModuleName f (Syntax.ModuleName _ unpacked) =
   Syntax.ModuleName () $ f unpacked
 
+
+modifyIdent :: (String -> String) -> Name -> Name
+modifyIdent f (Syntax.Ident l s) = Syntax.Ident l $ f s
+modifyIdent _ x = x
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
@@ -166,9 +166,6 @@ ihApp = foldl (Syntax.IHApp ())
 tyParen :: Type -> Type
 tyParen = Syntax.TyParen ()
 
-paren :: Exp -> Exp
-paren = Syntax.Paren ()
-
 type Match = Syntax.Match ()
 
 -- | A simple clause of a function binding.
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
@@ -287,7 +287,9 @@ generateEnumDecls Proto3 info =
     --         = Prelude.show k
     --       showEnum k = Prelude.show k
     --       readEnum "Enum2a" = Prelude.Just Enum2a -- alias
-    --       readEnum k = Text.Read.readMaybe k
+    --       readEnum "Enum2" = Prelude.Just Enum2
+    --       readEnum "Enum1" = Prelude.Just Enum1
+    --       readEnum k = Text.Read.readMaybe k >>= maybeToEnum
     , instDecl [] ("Data.ProtoLens.MessageEnum" `ihApp` [dataType])
         [ [ match "maybeToEnum" [pLitInt k] $ "Prelude.Just" @@ con (unQual c)
           | (c, k) <- constructorNumbers
@@ -317,7 +319,6 @@ generateEnumDecls Proto3 info =
           | v <- enumValues info
           , let n = enumValueName v
           , let pn = T.unpack $ enumValueDescriptor v ^. name
-          -- problem: descriptor != name, only name has Foo' prefix, not descriptor
           ] ++
           [match "readEnum" [pVar "k"] $ ("Prelude.>>=" @@ paren ("Text.Read.readMaybe" @@ "k")) @@ ("Data.ProtoLens.maybeToEnum")]
         ]
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
@@ -108,8 +108,8 @@ generateModule modName imports syntaxType modifyImport definitions importedEnv
               [ "Prelude", "Data.Int", "Data.Word"
               , "Data.ProtoLens", "Data.ProtoLens.Message.Enum"
               , "Lens.Family2", "Lens.Family2.Unchecked", "Data.Default.Class"
-              , "Data.Text",  "Data.Map" , "Data.ByteString"
-              , "Lens.Labels"
+              , "Data.Text",  "Data.Map", "Data.ByteString"
+              , "Lens.Labels", "Text.Read"
               ]
             ++ map importSimple imports
     env = Map.union (unqualifyEnv definitions) importedEnv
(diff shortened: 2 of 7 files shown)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
```diff
@@ -359,3 +359,7 @@ modifyModuleName :: (String -> String) -> ModuleName -> ModuleName
 modifyModuleName f (Syntax.ModuleName _ unpacked) =
   Syntax.ModuleName () $ f unpacked
 
+
+modifyIdent :: (String -> String) -> Name -> Name
```
judah: Code structure nit: we try to define names inside the Definitions module and use them in Generate.  In this case, adding new fields to `EnumInfo` with the `Name`s for the unrecognized types:
https://github.com/google/proto-lens/blob/master/proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs#L157

pkinsky:
I signed the CLA (by adding myself to the takt OSS google group), cc @judah

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -259,80 +259,185 @@ generateEnumExports syntaxType e = [exportAll n, exportWith n aliases]
...
+    --       maybeToEnum 3 = Prelude.Just Enum2
+    --       maybeToEnum k
+    --         = Prelude.Just
+    --             (FooEnum'Unrecognized
+    --               (FooEnum'UnrecognizedValue (Prelude.fromIntegral k)))
+    --       showEnum (FooEnum'Unrecognized (FooEnum'UnrecognizedValue k))
+    --         = Prelude.show k
+    --       showEnum k = Prelude.show k
+    --       readEnum "Enum2a" = Prelude.Just Enum2a -- alias
+    --       readEnum k = Text.Read.readMaybe k
```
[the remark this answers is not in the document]
pkinsky: `readMaybe k >>= maybeToEnum` assumes that the textual representation of these enum's is an Int. I was trying to reproduce the pre-existing logic in which strings like "Enum2a", "Enum2", etc are converted into instances of the sum type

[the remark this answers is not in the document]
judah: Oh, I see, I think I misread your code earlier.

Taking another look, we actually can't use the `Show` and `Read` instances for `FooEnum` to implement `showEnum/readEnum`.  Because for nested enums like
```
message Bar {
  enum Foo { A = 1; }
}
```
we want `showEnum Bar'A = "A"` and `readEnum "A" = Bar'A`, not `showEnum Bar'A = "Bar'A"` and `readEnum "Bar'A" = Bar'A`.  It could probably be described better in the docs, but `readEnum/showEnum` are specifically for the proto text format representation of the enum:
https://github.com/google/proto-lens/blob/master/proto-lens/src/Data/ProtoLens/Message.hs#L231

I think we should just take the string representations (*not* the ints like I suggested earlier), as in the code for proto2:
https://github.com/google/proto-lens/blob/master/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs#L374

```
readEnum "Enum2" = Enum2
readEnum "Enum2a" = Enum2a
etc.
```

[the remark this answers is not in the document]
pkinsky: That sounds good, but that still leaves us with no ability to serialize/deserialize unrecognized enum's. I'm assuming `readEnum . showEnum == id` is a property we want to maintain. If so what should `showEnum Foo'FooEnum Foo'Unrecognized...` result in?

proto-lens-tests/tests/proto3.proto
```diff
@@ -16,17 +16,18 @@ message Foo {
   }
   Sub sub = 5;
 
-  enum FooEnum {
```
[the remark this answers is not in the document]
pkinsky: I think that was due to a transient issue, I've reverted this change.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -259,80 +259,185 @@ generateEnumExports syntaxType e = [exportAll n, exportWith n aliases]
...
+
+    defaultCon = con $ unQual $ head constructorNames
+
+    minBoundName = head constructorNames
+    maxBoundName = last constructorNames
+
+    constructorNumbers = map (second (fromIntegral . (^. number))) constructors
+
+    succPairs = zip constructorNames $ tail constructorNames
+    succDecl funName boundName thePairs =
```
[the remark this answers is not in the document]
pkinsky: Sounds good. Is not erroring on `succ maxEnum` and instead returning `SomeEnum'UnrecognizedValue... (fromEnum maxEnum + 1`) the desired behavior here?

Do you also want me to redefine `pred` as `toEnum . (-1) . fromEnum`? Or should `pred minEnum` still result in an error?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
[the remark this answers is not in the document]
judah: Ah, good point, we do need something sensible for `showEnum` at least.  How about this?
```
showEnum Enum2 = "Enum2"
showEnum (Foo'Unrecognized (Foo'UnrecognizedValue n)) = show n

readEnum "Enum2" = Enum2
readEnum "Enum2a" = Enum2
readEnum s = Foo'Unrecognized . UnrecognizedValue <$> readMaybe s
```

Since in proto text format, it's fine to refer to an enum by its numeric value rather than the case name.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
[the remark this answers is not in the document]
pkinsky: I think this would break things, eg for `FooEnum` as defined in proto3.proto:
```
  enum FooEnum {
    option allow_alias = true;
    Enum1 = 0;
    Enum2 = 3;
    Enum2a = 3;
  }
```
it seems like we want `succ Enum1 == Enum2`, not `succ Enum1 == Foo'FooEnum'Unrecognized (Foo'FooEnum'UnrecognizedValue 2)`

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -259,80 +259,193 @@ generateEnumExports syntaxType e = [exportAll n, exportWith n aliases]
...
-                $ "Prelude.show" @@ "k"
-          ]
-
         , [ match "readEnum" [stringPat pn]
               $ "Prelude.Just" @@ con (unQual n)
           | v <- enumValues info
           , let n = enumValueName v
           , let pn = T.unpack $ enumValueDescriptor v ^. name
+          ] ++
+          [match "readEnum" [pVar "k"] $ ("Prelude.>>=" @@ paren ("Text.Read.readMaybe" @@ "k")) @@ ("Data.ProtoLens.maybeToEnum")]
```
judah: Break this line which has gotten pretty long.

proto-lens-tests/src/Data/ProtoLens/TestUtil.hs
```diff
@@ -68,6 +68,7 @@ import Text.PrettyPrint
 testMain :: [Test] -> IO ()
 testMain = defaultMain
 
+-- todo use this for enum msg on Foo
```
judah: Remove comment

proto-lens-tests/tests/proto3_test.hs
```diff
@@ -98,16 +98,20 @@ main = testMain
       , testCase "string" $ (def :: Strings) @=? (def & string .~ "")
       , testCase "enum" $ (def :: Foo) @=? (def & enum .~ Foo'Enum1)
       ]
-  -- Enums are all pattern aliases
+  -- Enums are sum types, except for aliases
   , testGroup "enum"
       [ testCase "aliases are exported" $ Foo'Enum2 @=? Foo'Enum2a
+      , serializeTo "serializeTo enum"
```
judah: Can you add a copy of this test which uses an unrecognized value?

Also: replace l.107 with more explicit generation:
```
tagged 6 $ VarInt 3
```
(where `6` is the proto tag of that field, and `3` is the value corresponding to `Enum2`)

proto-lens-tests/tests/proto3_test.hs
```diff
@@ -98,16 +98,20 @@ main = testMain
...
+          "0\ETX"
       , testCase "enum values" $ do
           map toEnum [0, 3, 3] @=? [Foo'Enum1, Foo'Enum2, Foo'Enum2a]
-          ["Foo'Enum1", "Foo'Enum2", "Foo'Enum2", "toEnum 5"]
+          ["Foo'Enum1", "Foo'Enum2", "Foo'Enum2", "Foo'FooEnum'Unrecognized (Foo'FooEnum'UnrecognizedValue 5)"]
               @=? map show [Foo'Enum1, Foo'Enum2, Foo'Enum2a, toEnum 5]
           ["Enum1", "Enum2", "Enum2", "6"]
               @=? map showEnum [Foo'Enum1, Foo'Enum2, Foo'Enum2a, toEnum 6]
-          [Just Foo'Enum1, Just Foo'Enum2, Just Foo'Enum2, Nothing, Nothing]
+          [Just Foo'Enum1, Just Foo'Enum2, Just Foo'Enum2, maybeToEnum 4, maybeToEnum 5]
```
judah: Test nit (working around the fact that the constructor isn't provided): Add a new test that checks `fromEnum <$> maybeToEnum 4 == Just 4`

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -259,80 +259,193 @@ generateEnumExports syntaxType e = [exportAll n, exportWith n aliases]
...
+    -- instance Data.ProtoLens.MessageEnum FooEnum where
+    --       maybeToEnum 0 = Prelude.Just Enum1
+    --       maybeToEnum 3 = Prelude.Just Enum2
+    --       maybeToEnum k
+    --         = Prelude.Just
+    --             (FooEnum'Unrecognized
+    --               (FooEnum'UnrecognizedValue (Prelude.fromIntegral k)))
+    --       showEnum (FooEnum'Unrecognized (FooEnum'UnrecognizedValue k))
+    --         = Prelude.show k
+    --       showEnum k = Prelude.show k
```
judah: I think this comment needs to be updated since you're not calling `Prelude.show` anymore.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -259,80 +259,193 @@ generateEnumExports syntaxType e = [exportAll n, exportWith n aliases]
...
+        (  (flip conDecl [] <$> constructorNames)
+        ++ [conDecl unrecognizedName [tyCon $ unQual unrecognizedValueName]]
+        )
+        $ deriving' ["Prelude.Show", "Prelude.Eq", "Prelude.Ord, Prelude.Read"]
+
+    -- newtype FooEnum'UnrecognizedValue = FooEnum'UnrecognizedValue Data.Int.Int32
+    --   deriving (Prelude.Eq, Prelude.Ord, Prelude.Show, Prelude.Read)
+    , newtypeDecl unrecognizedValueName
+       "Data.Int.Int32"
+        $ deriving' ["Prelude.Eq", "Prelude.Ord", "Prelude.Show, Prelude.Read"]
```
judah: Can we remove the `Read` instance, to make sure it's not being used accidentally?

(Also because we'd like `toEnum/maybeToEnum` to be the only way to construct these values)

Up to you.

proto-lens-tests/tests/proto3_test.hs
judah: Ping on the first line of the above comment (unless I'm misreading the diff).

More specifically, please add this test which checks that unrecognized values round-trip through serialization as expected.
```
serializeTo "serializeTo enum unknown"
    (def & enum .~ toEnum 5 :: Foo)
    "enum: 5"
    $ tagged 6 $ VarInt 5
```

pkinsky: whoops, sorry about that. Fixed.

PR #154, opened 2017-11-14
Service Generation

isovector:
This PR adds support for `service` constructs in proto files. Sending your way for early thoughts on it.

---

```
service SearchService {
  rpc Search (Foo) returns (Bar);
}
```

becomes

```haskell
data SearchService = SearchService{}

instance Data.ProtoLens.Service.Types.Service SearchService where
        type ServiceName SearchService = "SearchService"
        type ServicePackage SearchService = "test.any"
        type ServiceMethods SearchService = '["search"]

instance Data.ProtoLens.Service.Types.HasMethodImpl SearchService
           "search"
         where
        type MethodName SearchService "search" = "Search"
        type MethodInput SearchService "search" = Foo
        type MethodOutput SearchService "search" = Bar
        type MethodStreamingType SearchService "search" = 'Data.ProtoLens.Service.Types.NonStreaming
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/154)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
```diff
@@ -253,6 +253,9 @@ unQual = Syntax.UnQual ()
 type TyVarBind = Syntax.TyVarBind ()
 type Type = Syntax.Type ()
 
+tyFun :: Type -> Type -> Type
```
isovector: note to self: unnecessary now

proto-lens-protoc/package.yaml
```diff
@@ -55,6 +55,7 @@ library:
     - Data.Default.Class as Data.ProtoLens.Reexport.Data.Default.Class,
     - Data.Map as Data.ProtoLens.Reexport.Data.Map,
     - Data.ProtoLens as Data.ProtoLens.Reexport.Data.ProtoLens,
+    - Data.ProtoLens.GRPC.Types as Data.ProtoLens.Reexport.Data.ProtoLens.GRPC.Types,
```
judah: Let's call this module `Data.ProtoLens.Service.Types`, since there are multiple users of the "service" spec (in particular Google's internal Stubby protocol).

proto-lens/package.yaml
```diff
@@ -4,11 +4,11 @@ synopsis: A lens-based implementation of protocol buffers in Haskell.
 description: >
   The proto-lens library provides to protocol buffers using modern
   Haskell language and library patterns.  Specifically, it provides:
-   
+
```
judah: Don't remove this whitespace; it messes up the comments due to interactions between Haddock/cabal/hpack.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -202,6 +242,30 @@ collectDefinitions fd = let
...
                           (fd ^. messageType) (fd ^. enumType)
 
+collectServices :: FileDescriptorProto -> [ServiceInfo]
+collectServices fd = fmap toServiceInfo $ fd ^. service
+  where
+    toServiceInfo :: ServiceDescriptorProto -> ServiceInfo
+    toServiceInfo sd =
+        ServiceInfo
+            { serviceName = fromString $ T.unpack $ sd ^. name
+            , serviceMethods = fmap (toMethodInfo . (<> "'") . lowerInitialChars $ sd ^. name) $ sd ^. method
```
judah: Use `camelCase` instead of `lowerInitialChars`?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -191,6 +224,13 @@ definedFieldType fd env = fromMaybe err $ Map.lookup (fd ^. typeName) env
     err = error $ "definedFieldType: Field type " ++ unpack (fd ^. typeName)
                   ++ " not found in environment."
 
+-- | Look up the Haskell name for the type of a given type.
+definedType :: Text -> Env QName -> Definition QName
```
judah: Code structure: Do all name resolution within `Definition`, rather than deferring it to `Generate`.  I.e., `ServiceInfo` should contain the Haskell names of the types, not just the proto names, similar to what we do for (sub)messages.  For example, by passing an Env to `collectServices`.

(Also: merge this with the function at l.221)

proto-lens/src/Data/ProtoLens/GRPC/Types.hs
```diff
@@ -0,0 +1,21 @@
...
+type ServerRWHandler     a b = a -> IO (Maybe a) -> (b -> IO ()) -> IO ()
+
+data SomeHandler where
+  NormalHandler :: (Message a, Message b) => ServerHandler       a b -> SomeHandler
+  ReaderHandler :: (Message a, Message b) => ServerReaderHandler a b -> SomeHandler
+  WriterHandler :: (Message a, Message b) => ServerWriterHandler a b -> SomeHandler
+  BiDiHandler   :: (Message a, Message b) => ServerRWHandler     a b -> SomeHandler
+
+class Service s where
+  packHandlers :: s -> [SomeHandler]
```
judah: Can this also support client-side RPC calls?  Or would we need another method?

proto-lens/package.yaml
isovector: that's autosave :(

proto-lens/src/Data/ProtoLens/GRPC/Types.hs
isovector: there are client-side rpcs?

judah: I mean for client-side code that sends requests to gRPC servers.  The `packHandlers` method seems more suited for implementing the gRPC servers themselves.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
isovector: How strongly do you feel about this? `hsFieldType` and friends already do this in `Generate`, and there is no unified `Env` to lookup until `Generate.`

proto-lens/src/Data/ProtoLens/GRPC/Types.hs
isovector: good question. do you have a mental model of what the client-side interface should look like? in particular the streaming interfaces.

isovector:
@judah client-side of RPCs chat got buried due to github being stupid. do you have an idea about what a client-side streaming interface should look like?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
judah: You're right, my bad.  The way you wrote it SGTM.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
```diff
@@ -259,9 +289,15 @@ tyCon = Syntax.TyCon ()
 tyList :: Type -> Type
 tyList = Syntax.TyList ()
 
+tyPromotedList :: [Type] -> Type
+tyPromotedList ts = Syntax.TyPromoted () $ Syntax.PromotedList () True ts
+
 tyPromotedString :: String -> Type
 tyPromotedString s = Syntax.TyPromoted () $ Syntax.PromotedString () s s
 
+tyPromotedBool :: Bool -> Type
```
judah: Style nit: specializing to `Bool` feels off for this file.  Instead:
```
type Promoted = Syntax.Promoted ()
tyPromoted :: Promoted -> Type
tyPromoted = Syntax.TyPromoted ()

instance FromString Promoted where
    fromString = Syntax.PromotedCon () True
```
and then use `tyPromoted "'True"`, etc. in Generate.

proto-lens/package.yaml
```diff
@@ -4,11 +4,11 @@ synopsis: A lens-based implementation of protocol buffers in Haskell.
...
```
judah: Revert the whitespace changes to this file.

proto-lens/src/Data/ProtoLens/Service/Types.hs
```diff
@@ -0,0 +1,40 @@
...
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE GADTs                 #-}
+{-# LANGUAGE KindSignatures        #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+
+module Data.ProtoLens.Service.Types where
```
judah: Module-level haddock comment.

Also, use an explicit export list.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -97,6 +110,20 @@ data MessageInfo n = MessageInfo
...
     } deriving Functor
 
+data ServiceInfo = ServiceInfo
+    { serviceName :: Text
+    , serviceMethods :: [MethodInfo]
+    }
+
+data MethodInfo = MethodInfo
+    { methodIdent :: Text
+    , methodPath :: Text
```
judah: I think this path structure ("/pkg.Foo/Bar") is specific to gRPC rather than being part of the official proto spec.  I suggest we just record the name of  the method ("Bar") and also the name of the service ("pkg.Foo") in a new field of `ServiceInfo`.)

proto-lens-tests/tests/any.proto
```diff
@@ -10,3 +10,8 @@ message Bar {
   int32 a = 1;
   string b = 2;
 }
+
```
judah: Move this to a new unit test `service_test` with corresponding file `service.proto`.

proto-lens/src/Data/ProtoLens/Service/Types.hs
```diff
@@ -0,0 +1,40 @@
...
+import Data.ProtoLens.Message (Message)
+
+-- | Reifies the fact that there is a 'HasMethod' instance for every symbol
+-- claimed by the 'ServiceMethods' associated type.
+class HasAllMethods s (ms :: [Symbol])
+instance HasAllMethods s '[]
+instance (HasAllMethods s ms, HasMethod s m) => HasAllMethods s (m ': ms)
+
+
+class HasAllMethods s (ServiceMethods s) => Service s where
```
judah: Haddocks for the classes and methods.

proto-lens/src/Data/ProtoLens/Service/Types.hs
```diff
@@ -0,0 +1,40 @@
...
+  type ServiceName s    :: Symbol
+  type ServiceMethods s :: [Symbol]
+
+
+class ( KnownSymbol m
```
judah: Add `Service s` to the superclass constraints for this one?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
isovector: I'm not sure I agree with this -- literally just above this is code for promoting strings. Furthermore, the strings here are actually `Type`s, meaning if we wanted to pass the buck to calling code they'd need to do something like `tyPromoted $ fromString  $ show mybool` which is pretty disgusting imo. Thoughts?

stack-bootstrap.yaml
```diff
@@ -4,11 +4,11 @@ packages:
 # Build the current HEAD proto-lens-protoc against older revisions of proto-lens
 # and proto-lens-descriptors that are consistent with each other.
 - location:
-    commit: master
+    commit: 5766ceca2558b1721204ba74d3d080a3d98a7d37
```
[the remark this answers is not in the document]
isovector: It's necessary for bootstrapping the new `Service.Types` module until it exists upstream. Will revert just before merging.

isovector:
Addressed all comments -- ptal

blackgnezdo:
If we could get @awpr to review this it would be super-awesome. He has plenty of experience in the area.

Andrew, if you haven't seen, @isovector put together an excellent write-up about his journey: http://reasonablypolymorphic.com//blog/type-directed-code-generation

judah:
LGTM

isovector:
@blackgnezdo @awpr : I'm planning on merging this by the end of day unless you have substantial concerns with the approach!

blackgnezdo:
@awpr is OOO, we can polish this up later if need be.

judah:
@blackgnezdo, it looks like travis-ci has blacklisted this PR: https://travis-ci.org/google/proto-lens/requests

Is there an internal contact that you can ping to try to unblock this PR?  Otherwise we can try contacting TravisCI support directly I guess...

blackgnezdo:
I'm afraid I know nothing about Travis CI. @judah set it  up...

PR #157, opened 2017-11-28
Add function instance for IsLabel

thumphries:
This adds an orphan instance allowing protolens-generated labels to be used in function position.

Duplicate PR to bypass the Travis spam filter.

cc @judah

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/157)

### DIFF
--- a/lens-labels/Changelog.md
+++ b/lens-labels/Changelog.md
@@ -4,6 +4,8 @@
 - Improve readability of `HasLens` instances. (#118)
 - Remove support for `ghc-7.10`. (#136)
 - Use a `.cabal` file that's auto-generated from `hpack`. (#138)
+- Add an `IsLabel` instance for `(->)` allowing the use of labels
+  in function position. (#157)
 
 ## v0.1.0.2
 - Bump the dependency on `base` to support `ghc-8.2.1`.
--- a/lens-labels/Changelog.md
+++ b/lens-labels/Changelog.md
@@ -4,8 +4,8 @@
 - Improve readability of `HasLens` instances. (#118)
 - Remove support for `ghc-7.10`. (#136)
 - Use a `.cabal` file that's auto-generated from `hpack`. (#138)
-- Add an `IsLabel` instance for `(->)` allowing the use of labels
-  in function position. (#157)
+- Add `Lens.Labels.Unwrapped`, allowing easier interoperation with
+  other lens libraries. (#157)
 
 ## v0.1.0.2
 - Bump the dependency on `base` to support `ghc-8.2.1`.
--- a/lens-labels/package.yaml
+++ b/lens-labels/package.yaml
@@ -17,6 +17,7 @@ library:
   source-dirs: src
   exposed-modules:
     - Lens.Labels
+    - Lens.Labels.Unwrapped
   dependencies:
     - base >= 4.8 && < 4.11
     - ghc-prim >= 0.4 && < 0.6
--- a/lens-labels/src/Lens/Labels.hs
+++ b/lens-labels/src/Lens/Labels.hs
@@ -52,6 +52,7 @@ module Lens.Labels (
     view,
     ) where
 
+
 import qualified Control.Category as Category
 import GHC.Prim (Proxy#, proxy#)
 import GHC.OverloadedLabels (IsLabel(..))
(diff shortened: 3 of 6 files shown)


thumphries:
This should probably lead to a version bump? (added exported entity)

judah:
The current Cabal versions in HEAD haven't been released yet; i.e., the version bump has already occurred.  You can just update the changelog for `lens-labels-0.2.0.0` here: https://github.com/google/proto-lens/blob/master/lens-labels/Changelog.md

lens-labels/src/Lens/Labels/Function.hs
```diff
@@ -0,0 +1,16 @@
...
+------------------------------------------------------------------------------
+-- | This module provides OverloadedLabels 'IsLabel' support to lenses
+-- generated by protolens; eg, it allows you to reference a proto field 'foo'
+-- as @#foo@ without having protolens steal perfectly good identifier names.
+module Lens.Labels.Function where
+
+import GHC.OverloadedLabels (IsLabel (..))
+import Lens.Labels (LensFn (..))
+
+instance IsLabel x (LensFn p q) => IsLabel x (p -> q) where
```
judah: Looks like this breaks on 8.2.1 due to a change to IsLabel:
https://travis-ci.org/google/proto-lens/jobs/308690066

You can copy the CPP logic from here: https://github.com/google/proto-lens/blob/master/lens-labels/src/Lens/Labels.hs#L83

And test whether that fixed it with `stack test --pedantic --resolver=nightly-2017-07-31`

lens-labels/src/Lens/Labels/Function.hs
```diff
@@ -0,0 +1,16 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+------------------------------------------------------------------------------
+-- | This module provides OverloadedLabels 'IsLabel' support to lenses
```
judah: Maybe mention the word "orphan" in this module description?  Up to you.

lens-labels/src/Lens/Labels/Function.hs
```diff
@@ -0,0 +1,16 @@
...
+-- generated by protolens; eg, it allows you to reference a proto field 'foo'
+-- as @#foo@ without having protolens steal perfectly good identifier names.
```
judah: Nit: theoretically this package isn't specific to `proto-lens`, so maybe don't mention protos in the comment (just "records" more generally)?

thumphries: Whoops, just copied your comment across without thinking

thumphries:
`-Werror` is complaining about "fragile instances":

```

    /Users/tim/src/proto-lens/lens-labels/src/Lens/Labels/Function.hs:21:10: warning: [-Wsimplifiable-class-constraints]
        • The constraint ‘IsLabel x (LensFn p q)’
            matches an instance declaration
          instance (p ~ (a -> f b), q ~ (s -> f t),
                    Lens.Labels.HasLens f s t x a b) =>
                   IsLabel x (LensFn p q)
            -- Defined at src/Lens/Labels.hs:82:5
          This makes type inference for inner bindings fragile;
            either use MonoLocalBinds, or simplify it using the instance
        • In the context: IsLabel x (LensFn p q)
          While checking an instance declaration
          In the instance declaration for ‘IsLabel x (p -> q)’
       |
    21 | instance IsLabel x (LensFn p q) => IsLabel x (p -> q) where
       |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

... I've never seen this warning before, I'm not quite sure if it is useful or not.

judah:
Ah...I suppose the point of that warning is we *could* simplify things by just copying the instance from here, but with raw functions instead of `LensFn`, rather than using the superclass interaction: https://github.com/google/proto-lens/blob/master/lens-labels/src/Lens/Labels.hs#L80

Honestly though I wouldn't mind just disabling that warning in an `OPTIONS_GHC` pragma for this module.  If the relevant test ([labels_test](https://github.com/google/proto-lens/blob/master/proto-lens-tests/tests/labels_test.hs)) still passes, then it should be pretty safe.

Up to you which approach to take.    For context this appears to be the ticket that caused introduction of the new warning: https://ghc.haskell.org/trac/ghc/ticket/11948

thumphries:
- Added a test
- Updated documentation
- Renamed the new module `Unwrapped` 

I came around to your point of view regarding the name. This is a module that you'd reach for when frustrated by the wrapping/unwrapping, and the fact it introduces an instance for `(->)` is probably not something the user will care about.

PR #159, opened 2017-12-08
Expose shrinkMessage

thumphries:
This makes it easier to call `forAllShrink`, which is what you're stuck with if you forego QuickCheck's TH property discovery. Currently this is only accessible via the `ArbitraryMessage` newtype.

Before:

```haskell
forAllShrink arbitraryMessage (fmap unArbitraryMessage . shrink . ArbitraryMessage)
```

After:

```haskell
forAllShrink arbitraryMessage shrinkMessage
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/159)


judah:
Makes sense, thanks!

PR #163, opened 2017-12-15
Better arbitrary generator

FintanH:
After discovering that recursive proto data can cause the arbitrary generator to create large structures which in turn causes memory blowout. 

The fix is to reduce the `size` of generated structures on each recursive call.

To test this I also added `pathological.proto` an offending data structure to the case, and a test `pathological_test.hs` to test round trip serialization of the proto data.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/163)


judah:
LGTM; I suspect more complicated protos (e.g. a recursive tree structure with larger branching factor) will still run into this issue, but this seems like a good workaround for now.

PR #166, opened 2018-01-07
Add protobuf timestamp support

### LINKED ISSUE #165: google/protobuf/timestamp.proto support
Is the omission of the predefined timestamp proto intentional?

TPolzer:
Fix #165. I also have functions for Timestamp <-> UTCTime conversion lying around but was unsure whether they should be in this repo and under which package.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/166)

### DIFF
--- a/proto-lens-protobuf-types/package.yaml
+++ b/proto-lens-protobuf-types/package.yaml
@@ -15,6 +15,7 @@ extra-source-files:
   - proto-src/google/protobuf/any.proto
   - proto-src/google/protobuf/duration.proto
   - proto-src/google/protobuf/wrappers.proto
+  - proto-src/google/protobuf/timestamp.proto
 
 custom-setup:
   dependencies:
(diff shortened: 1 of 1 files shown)


TPolzer:
The CLA message is strange, I put my @google.com address in the commit (even though this is my spare time obviously).

judah:
Thanks for the patch!  This change LGTM.

The last time I looked at it, it seemed like it's not exactly "correct" (or at least, not easy) to convert between `timestamp.proto` and Haskell's `UTCTime` due to their different treatments of leap seconds:
https://github.com/google/protobuf/blob/master/src/google/protobuf/timestamp.proto#L43
https://hackage.haskell.org/package/time-1.8.0.4/docs/Data-Time-Clock.html
So at the moment I'm cautious about including such conversion functions in proto-lens proper.

For what it's worth, in a recent project we worked around this inconsistency by creating a new custom message type with closer semantics to `UTCTime`.  Maybe the conversions better off as a separate package and/or repo for now?  Let me know if you have any thoughts.

PR #170, opened 2018-01-16
Build with `haskell-src-exts-1.20.*`.

judah:
Tested with `stack test --resolver=nightly-2018-01-15`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/170)

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
@@ -67,30 +67,41 @@ recDecl dataName fields
 type Decl = Syntax.Decl ()
 
 patSynSig :: Name -> Type -> Decl
+#if MIN_VERSION_haskell_src_exts(1,20,0)
+patSynSig n t = Syntax.PatSynSig () [n] Nothing Nothing Nothing t
+#else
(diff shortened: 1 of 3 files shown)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Combinators.hs
```diff
@@ -67,30 +67,41 @@ recDecl dataName fields
 type Decl = Syntax.Decl ()
 
 patSynSig :: Name -> Type -> Decl
+#if MIN_VERSION_haskell_src_exts(1,20,0)
```
blackgnezdo: I'm curious if anybody ran analysis on all Hackage for MIN_VERSION CPP gunk, which package would come out on top?

judah: In all likelihood, probably the `base` package itself :p
https://www.google.com/search?ei=7IleWufIDOPU0gLbj7Ug&q=%22min_version_base%22+site%3Ahackage.haskell.org&oq=%22min_version_base%22+site%3Ahackage.haskell.org&gs_l=psy-ab.3...6405.9216.0.9445.7.6.0.0.0.0.106.549.4j2.6.0....0...1c.1.64.psy-ab..2.0.0....0.x18W9bQot7A

PR #172, opened 2018-01-19
Field comments

clavoie-takt:
Revive @judah 's old field-comments branch, and prettify the output a wee bit.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/172)

Issue #167, opened 2018-01-07
Should generated types derive Generic by default?


TPolzer:
Very useful for JSON parsing etc..

blackgnezdo:
These should be easy enough to derive as orphan instances. I don't think burdening the compiler with Generic by default is a good trade-off.

TPolzer:
Tbh I only learned about StandaloneDeriving today. Unfortunately it doesn't work 100% because the constructors for generated enum's UnrecognizedValue are not exported. Is dd2a6e257d63f8b73422b46aa241d6c60e3619da enough or is there more that I just didn't try to use?

blackgnezdo:
I'll let @judah decide if exporting everything is fine. Another alternative is to make derive Generic be a [custom message level option](https://developers.google.com/protocol-buffers/docs/proto#options). Something like 

```
message MyMessage {
  option (haskell_deriving) = "Generic";
}
```

Not sure what's the most ergonomic options type though.

judah:
It's intentional that we *don't* export the UnrecognizedValue constructor, since using it blindly can lead to inconsistent code (e.g. violating equality laws).

I agree let's be conservative and not derive Generic by default to avoid slower compilation times.  I'd be fine with an `option(haskell_deriving)`, though implementing that in `proto-lens-protoc` would be blocked on #27.

If you do specifically want JSON parsing, going through Generic is probably the wrong approach since the (Haskellized, camel-cased) field names won't match up with the proto names.  See #128 instead.

vrom911:
I'm using standalone deriving for writing instances for generated data types.
But it requires from me to make `deriving instance Generic WireValue` but I see the following error:
```
error:
    • Can't make a derived instance of ‘Generic WireValue’:
        WireValue must be a vanilla data constructor
    • In the stand-alone deriving instance for ‘Generic WireValue’
   |
46 | deriving instance Generic WireValue
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
So what should I do?

ChShersh:
I understand that deriving `Generic` instances by default will slowdown compilation. Having custom message level option works fine to me!

PR #174, opened 2018-01-30
Hide the constructors and raw record fields of message types.

judah:
They're not that useful for constructing values (especially after adding the
"unknown fields" case) and were previously mostly used for documentation.  Now
that #172 has landed, we have a better source of documentation that doesn't
expose internal details.

Also change the `Show` instance for messages to be a wrapper around
`showMessageShort`.  This both hides the now-inaccessible record fields and
shortens the output in most cases.  For example:

```
> print (def :: DescriptorProto)
{}
> print (def & #name .~ "hello" & #field .~ [def, def & #name .~ "bye"]:: DescriptorProto)
{name: "hello" field { } field { name: "bye" }}
```

Also bump the resolver for bootstrapping to match the regular build.

Hide message constructors.

showsPrec, and bootstrap

### DIFF
--- a/stack-bootstrap.yaml
+++ b/stack-bootstrap.yaml
@@ -1,4 +1,4 @@
-resolver: lts-9.21
+resolver: lts-10.3
 packages:
 - proto-lens-protoc
 # Build the current HEAD proto-lens-protoc against older revisions of proto-lens
--- a/proto-lens-protoc/Changelog.md
+++ b/proto-lens-protoc/Changelog.md
@@ -9,12 +9,14 @@
 - Don't generate Haskell modules if they won't be used. (#126)
 - Bundle enum pattern synonyms exports with their type. (#136)
 - Split the `Message` class into separate methods. (#139)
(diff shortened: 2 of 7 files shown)


stack-bootstrap.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-9.21
+resolver: lts-10.3
```
blackgnezdo: Does this foreclose any bootstrap options?

judah: Not really; as long as we keep testing the older LTS's in TravisCI, we'll be reasonable sure it's possible to bootstrap with older versions by manually editing this file.  I think it's just more convenient for developers if this is in sync with `stack.yaml`.

judah:
Unfortunately this causes problems related to #167; since the types don't have exposed constructors, users can't derive their own instances for them.

For example: https://travis-ci.org/google/proto-lens/jobs/335018254

Leaving open for now while I try to think of a workaround.

PR #177, opened 2018-02-08
Don't unnecessarily touch files. Fixes #176.

### LINKED ISSUE #176: Haskell files being touched causes unecessary recompilation
`proto-lens-protoc` touches (updates the modification time of) generated `.hs` files even when the contents did not change, causing unecessary rebuilds when compiling with ghc.

This is because currently GHC considers only the mtime of the input file for determining whether something has to be recompiled, not its contents.

The problematic code is here: https://github.com/google/proto-lens/blob/fe05638fbe927d1becf7a2eabef6fad71ed0768a/proto-lens-protoc/src/Data/ProtoLens/Setup.hs#L307-L312

It seems that it is `protoc` that touches / rewrites the files.

Which way should this be fixed?

* Generate the files into a different, temporary `output` directory, the moving the `.hs` files over only if they aren't identical?
* Change `protoc` itself to not write the files if the contents are the same?

nh2:
See #176.

This avoids unnecessary recompilation by GHC in projects
that use proto-lens-protoc.

Avoiding to touch the `.hs` files is the important bit.

The avoidance of touching `.proto` files is not necessary
according to my knowledge, avoiding writing the `.hs` is
technically enough for the GHC use case, but in cases
where the user does some similar recompilation avoidance
logic with the copied `.proto` files, this will make their
life easier.


proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -209,10 +216,41 @@ generateSources root l files = do
...
+    -- Discover generated files.
+    -- `getDirectoryContentsRecursive` is lazy IO, so we `force` through
+    -- the list to make it strict hereinafter.
+    !generatedFiles <- force <$> getDirectoryContentsRecursive tmpAutogenModulesDir
+    -- Move files to autogen dir only if file contents are different.
+    forM_ generatedFiles $ \pathRelativeToTmpDir -> do
+        let sourcePath = tmpAutogenModulesDir </> pathRelativeToTmpDir
+        let targetPath = autogenModulesDir l </> pathRelativeToTmpDir
+        identical <- do
+            targetExists <- doesFileExist targetPath
```
blackgnezdo: What do you think about comparing file sizes before doing the full read? If you don't think it's worth the trouble because the common case is the files are identical, maybe mention that in the comment?

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -209,10 +216,41 @@ generateSources root l files = do
...
+    let tmpAutogenModulesDir = autogenModulesDir l ++ "-protoc-tmpOutDir"
+    -- Generate .hs files into temp dir.
+    generateProtosWithImports (root : importDirs) tmpAutogenModulesDir
                               -- Applying 'root </>' does nothing if the path is already
                               -- absolute.
                               (map (root </>) files)
+    -- Discover generated files.
+    -- `getDirectoryContentsRecursive` is lazy IO, so we `force` through
+    -- the list to make it strict hereinafter.
+    !generatedFiles <- force <$> getDirectoryContentsRecursive tmpAutogenModulesDir
```
blackgnezdo: I am curious if you do this out of general principle that lazy IO is bad or because there's an observable difference.

nh2: The former (to make for least-surprising code)

blackgnezdo:
LGTM

@judah are you OK with this change?

nh2:
@blackgnezdo One sec, still answering the other question :D

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
nh2: @blackgnezdo I could do that, but I'd have to do it using [`withFile path ReadMode hFileSize`](https://stackoverflow.com/questions/5620332/what-is-the-best-way-to-retrieve-the-size-of-a-file-in-haskell/5621159#5621159), or add a dependency on [`unix-compat`](https://hackage.haskell.org/package/unix-compat-0.5.0.1/docs/System-PosixCompat-Files.html#v:fileSize) to [have it work on Windows while still being done done with `stat()` on other platforms](https://www.reddit.com/r/haskell/comments/3l8msi/is_there_a_portable_way_to_get_file_size/cv45hgy/).

In general comparing file size to eschew a read for only `stat` and `mv` is a good idea, but as you say, it's an optimisation for the case where the file does change, and in that case we have to recompile which dwarves the cost of the read, so it's your call on whether you think it justifies the extra logic. I have no opinion on it, but will happily add it or otherwise add a comment in this regard.

nh2:
OK, I answered https://github.com/google/proto-lens/pull/177#discussion_r167292928

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -233,7 +271,9 @@ copyProtosToDataDir verb root destDir files = do
         let destFile = destDir </> f
         createDirectoryIfMissingVerbose verb True
             (takeDirectory destFile)
-        installOrdinaryFile verb srcFile destFile
+        shouldCopy <- srcFile `moreRecentFile` destFile
```
judah: I don't think this check matters; can we remove it?  As far as I know, the installed data files don't trigger rebuilds the same way that the autogen source files are.

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
judah: I'm fairly certain it makes no difference here, here because the very next thing you do is call `forM_` on the result.

Normally it wouldn't matter much, but in this case you're adding a new dependency (`deepseq`) so I don't think it's worth it unless you've benchmarked it (or, perhaps, used the generated Core to proved that this way is better).

proto-lens-protoc/src/Data/ProtoLens/Setup.hs
```diff
@@ -209,10 +216,41 @@ generateSources root l files = do
                      [ InstalledPackageInfo.dataDir info </> protoLensImportsPrefix
                      | info <- collectDeps l
                      ]
-    generateProtosWithImports (root : importDirs) (autogenModulesDir l)
+    -- Generate .hs files into a temporary directory, then move them over
+    -- to the target (autogen) directory only if they are different from
+    -- what's already there. This way, we don't needlessly touch the generated
+    -- .hs files when nothing changes, and thus don't needlessly make GHC
+    -- recompile them (as it considers their modification times for that).
+    -- See https://github.com/google/proto-lens/issues/176
```
judah: Style nit: My preference is to avoid mentioning the issue number in the comment; it's redundant since we can rely on the other text (which you have written well) and/or on `git blame` for future readers to track this code back to its creation.

If you would still rather keep the issue number, that's fine, just reference it as `#176` rather than the full URL.

nh2:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[proto-lens-protoc/src/Data/ProtoLens/Setup.hs, line 224 at r1](https://reviewable.io:443/reviews/google/proto-lens/177#-L4qD5CbQRA5YnGiYm9R-r1-224:-L5FC1gqNdZM8QffF2ei:b-dho6d5) ([raw file](https://github.com/google/proto-lens/blob/ce954a5ec5cd2f7ebec0008f7ea615a4554da44d/proto-lens-protoc/src/Data/ProtoLens/Setup.hs#L224)):*
<details><summary><i>Previously, judah (Judah Jacobson) wrote…</i></summary><blockquote>

Style nit: My preference is to avoid mentioning the issue number in the comment; it's redundant since we can rely on the other text (which you have written well) and/or on `git blame` for future readers to track this code back to its creation.

If you would still rather keep the issue number, that's fine, just reference it as `#176` rather than the full URL.
</blockquote></details>

OK, I have removed it.

I'm not sure if the `git blame` reliance is a good argument, because blame info gets buried quickly when code is moved around or whitespace or typos are changed, and there aren't good tools to easily find the "original" commit that introduced a line across such modifications; so while certainly possible, it makes it unnecessarily harder. But I do buy the argument that the comment is descriptive enough and URL might be not be a stable target if the project moves somewhere else.

---

*[proto-lens-protoc/src/Data/ProtoLens/Setup.hs, line 234 at r1](https://reviewable.io:443/reviews/google/proto-lens/177#-L4qD5CbQRA5YnGiYm9R-r1-234:-L5F7-voTfbR-qdRbjD5:b-f5zbhv) ([raw file](https://github.com/google/proto-lens/blob/ce954a5ec5cd2f7ebec0008f7ea615a4554da44d/proto-lens-protoc/src/Data/ProtoLens/Setup.hs#L234)):*
<details><summary><i>Previously, judah (Judah Jacobson) wrote…</i></summary><blockquote>

I'm fairly certain it makes no difference here, here because the very next thing you do is call `forM_` on the result.

Normally it wouldn't matter much, but in this case you're adding a new dependency (`deepseq`) so I don't think it's worth it unless you've benchmarked it (or, perhaps, used the generated Core to proved that this way is better).
</blockquote></details>

Yes, the `forM_` I do afterwards forces it just as well. I didn't have performance in mind at all when I wrote this, only avoiding bugs. If somebody changes the code within the `forM_` in the future, they might trip over the lazy IO in the future.

To make a concrete example: If somebody did a(n accidentally) wrong `mv` in iteration `3` of the `forM_`, they can magically remove what would have been iteration `4` from the list, resulting in a form of "time travel" programming. Such issues take forever to debug (I spent a day on it once).

Thus I would heavily recommend against letting lazy IO creep into loops -- not for my own change, but for the sake of those who modify it after me. I considered adding a `deepseq` dependency not a big deal, because other `proto-lens` packages already depend on it (and I copied the version contraint from them), as does `Cabal`, on which `proto-lens-protoc` depends directly. If adding the dependency is undesired nevertheless, I could also drop in a small `getDirectoryContentsRecursiveStrict` helper that forces the lazy IO away before returning `generatedFiles`.

---

*[proto-lens-protoc/src/Data/ProtoLens/Setup.hs, line 274 at r1](https://reviewable.io:443/reviews/google/proto-lens/177#-L4qD5CbQRA5YnGiYm9R-r1-274:-L5FAWEhRVCdSYtgU3hG:brqedca) ([raw file](https://github.com/google/proto-lens/blob/ce954a5ec5cd2f7ebec0008f7ea615a4554da44d/proto-lens-protoc/src/Data/ProtoLens/Setup.hs#L274)):*
<details><summary><i>Previously, judah (Judah Jacobson) wrote…</i></summary><blockquote>

I don't think this check matters; can we remove it?  As far as I know, the installed data files don't trigger rebuilds the same way that the autogen source files do.
</blockquote></details>

I think you are correct with that (see also the commit message), but I have not yet tested if indeed it's not required to achieve our goal.

I kept it because in general, less mtime changes in my opinion (e.g. if some other form of build system picks up these files later), and this check is "free" as compared to the other one, it involves only stat() data and does not require reading the files for comparison. But it doesn't matter much to me, I can also remove it depending on your preference.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/proto-lens/177)*

nh2:
Short ping about this (not sure if my comments in Reviewable sent a ping)

judah:
Apologies for the delay in reply.  Let's remove the data-file mtime check since we don't have a concrete case where it'd be useful yet.  (And I could see build systems other than Cabal ending up just reimplementing the logic in this module themselves.)

Other than that, though, this change looks good me.

nh2:
> Let's remove the data-file mtime check since we don't have a concrete case where it'd be useful yet. (And I could see build systems other than Cabal ending up just reimplementing the logic in this module themselves.)

OK, I've pushed it out.

PR #180, opened 2018-03-05
Update Changelogs.

judah:
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/180)


judah:
Update: I've released new versions of the packages based on the latest master, and pushed tags for all the packages.

I've also deprecated proto-lens-descriptors on Hackage since it's now distributed as part of proto-lens.

PR #190, opened 2018-05-02
Fix README in Person Example

zliu41:
It previously contains some outdated stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/190)


judah:
Thank you for the patch.  The extra `Fields` module is actually for a *newer* version of `proto-lens` (`0.3.*`).  I've just sent a patch to add it back to Stackage here; apologies for the delay: https://github.com/fpco/stackage/pull/3593

Once that patch lands and proto-lens is back in Stackage, I suggest we update the tutorial to use the relevant nightly release.

zliu41:
SGTM

Issue #185, opened 2018-03-17
Support/use/document Cabal-2's autogen-modules


judah:
Cabal 2.0 supports an `autogen-modules` field which would improve our integration with `cabal sdist`: https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=autogen%20modules#autogenerated-modules

I'm not sure how that will interact with older versions of Cabal/stack, though.  It's also not yet available in a version of hpack that was released with stack.

PR #193, opened 2018-05-09
Fix #171 and bump to proto-lens-protoc-0.3.1.1.

### LINKED ISSUE #171: Use `autogenPackageModulesDir`
Cabal 2.0 added a function `autogenPackageModulesDir` which we should use in `Data.ProtoLens.Setup` if it's available.  That would let us generate modules separately for each component (e.g. library vs exe vs tests), rather than generating them all in one place.

At minimum, this would prevent confusing GHC/Cabal errors when an exe imports a proto module but doesn't specify it in other-deps, and the module *is* specified for a library, but the test accidentally doesn't depend on the library.

judah:
With `Cabal-2.0` and later, put the autogenerated modules of each component
(library, executables, tests, benchmarks) in their own directory.  This
eliminates some confusing warnings and/or build errors.

(Note: this required a change to the release script for generating sdists;
long-term we should replace it with something like #185.)

With `Cabal-1.*`, continue to put all autogenerated modules together
in the same directory.

An example of the previous issue, when an executable depended on a library
in the same `.cabal` file and the library provided proto modules:
```
<no location info>: warning: [-Wmissing-home-modules]
    These modules are needed for compilation but not listed in your .cabal
    file's other-modules: Proto.Lib
```


nhibberd:
LGTM

PR #196, opened 2018-05-21
Fix #185: Support and use autogen-modules.

### LINKED ISSUE #185: Support/use/document Cabal-2's autogen-modules

judah:
This makes it possible to generate tarballs without generated modules
(which would be regenerated anyway when the package is build).
See tensorflow/haskell#180 for an example of the issues that causes.

For `Cabal-1.*`, this continues the behavior as before.

Unfortunately, `hpack` requires `cabal-version: >=2.0` when you use
its `generated-modules` or `generated-other-modules` fields.
Our current set of LTSes that we support still includes `Cabal-1.*`
(which I think is correct).  Luckily we could work around that
using `hpack`'s `verbatim` field to accomplish the same thing a little
more verbosely.  Additionally, I mitigated the situation a little by changing
the Cabal test script to not `sdist` packages that we're not releasing
(`proto-lens-{tests/benchmarks}`).

This change required bumping `stack` to `1.7.1` in order to get new enough
versions of `Cabal` and `hpack`.  Happily, it greatly simplifies
the steps for releasing our packages.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/196)


blackgnezdo:
Please rebase?

PR #202, opened 2018-06-21
Separate compiler and run-time for sake of cross.

Ericson2314:
The run-time was pulled out of the compiler for sake for cross compilation. In that case, the compiler and run time are built for different platforms, so they need to be separate packages (until Cabal is improved).

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/202)

### DIFF
--- a/proto-lens-combinators/package.yaml
+++ b/proto-lens-combinators/package.yaml
@@ -20,7 +20,7 @@ custom-setup:
 
 dependencies:
   - base >= 4.9 && < 4.12
-  - proto-lens-protoc == 0.3.*
+  - proto-lens-protoc-rt == 0.3.*
   - lens-family == 1.2.*
 
 library:
@@ -42,7 +42,7 @@ tests:
       - lens-family-core
       - proto-lens
       - proto-lens-combinators
-      - proto-lens-protoc
+      - proto-lens-protoc-rt
       - test-framework
       - test-framework-hunit
     other-modules:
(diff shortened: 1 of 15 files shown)


blackgnezdo:
@judah what do you think?

@Ericson2314 you need to sign the CLA for the PR to proceed.

judah:
@Ericson2314 in addition to signing the CLA, can you you explain (or link to) how you're doing cross-compilation? I'm not familiar with such a setup or what the limitations of Cabal are (and Google wasn't an immediate help).  For example, what happens if proto-lens-protoc and another package in the build dependencies both depend on the same (transitive) library?

I'm not wild about increasing the number of Cabal packages we have to maintain, and would also like to think carefully about what to call these packages (for example, maybe `proto-lens-runtime` rather than `proto-lens-protoc-rt`).  On the face of it, though, organizing packages like this is not unreasonable.  (I also suspect this separation could also help simplify the Bazel integration story.)

Ericson2314:
I'm doing cross compilation with Nix. The nixpkgs Haskell cross compilation stuff is all public but my specific use of this is proprietary, sorry. (After this PR I can upstream it into nixpkgs however :)).

The limitation of Cabal is that all components of a package must be built with the same compiler. But the protoc setup module and runtime should be built with different compilers (e.g. GHC targeting the build machine for the Setup.hs and it's dependencies, and GHC/GHCJS targeting the host platform for the runtime.

There is no issue if the same library is needed by both. In the native case both will use the same build. In the cross case they will use different builds for different platforms.

judah:
As discussed offline with @Ericson2314, I'm closing this PR and redoing it in #203.  I confirmed with him that the work was done for Formation, which has already signed the Google CLA.

I ended up splitting into three packages instead of two, but I don't think that should cause any problems for your Nix setup.

PR #203, opened 2018-07-15
Split out proto-lens-protoc into three packages.

judah:
- `proto-lens-protoc`: Just the binary, along with internal/plugin modules
  for that binary.
- `proto-lens-runtime`: Support "reexported" modules that the generated
  files depend on.
- `proto-lens-setup`: Provides the custom Setup script
  (`Data.ProtoLens.Setup`).  Depends on the `proto-lens-protoc`.

It's annoying to have to maintain more packages.  However, this is probably
the right organization, for a couple reasons:

- As described in #202, this makes cross-compilation simpler due to some
  limitations of Cabal.
- Translating proto-lens into Bazel rules is less cumbersome because
  it naturally lends itself to split along these
  dimensions.  That is, the compiler and the runtime library are separate
  targets, and we don't need `Data.ProtoLens.`Setup` at all.

This causes a major version bump for `proto-lens-protoc` (to `0.4`)
and a minor version bump for packages that generate protos.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/203)


proto-lens-runtime/package.yaml
```diff
@@ -0,0 +1,48 @@
+name: proto-lens-runtime
+version: '0.4.0.0'
+author: Judah Jacobson
+maintainer: proto-lens@googlegroups.com
+github: google/proto-lens/proto-lens-runtime
+copyright: Google Inc.
+license: BSD3
+description: <
```
blackgnezdo: What does < do here vs > in some other description: > lines?

judah: It was a typo.  Fixed to `>`.

PR #204, opened 2018-07-27
Scalar type refs

pkinsky:
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/204)


judah:
Looks good!  Can you please also add a link to the docs index, so it shows up in `google.github.io/proto-lens`?
https://github.com/google/proto-lens/blob/master/docs/index.md

PR #206, opened 2018-08-08
Expose bindings for empty.proto.

judah:
We've found it useful to use that generic type, for example in RPCs that don't
return a value.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/206)


blackgnezdo:
LGTM

PR #208, opened 2018-08-21
Simplify lens-label instances.

judah:
One of proto-lens's biggest usability issues is the readability (or lack
thereof) of type error messages and type instances. We can improve things
somewhat by drastically simplifying the class we use for overloading labels.
The original version was made to be as general as possible, but no one's ever
used that generality in practice.

Before:
```
class HasLens (f :: * -> *) s t (x :: Symbol) a b where
   lensOf :: Proxy# x -> LensLike f s t x a b
```
to
```
class HasLens' s (x :: Symbol) a where
   lensOf' :: Functor f => Proxy# x -> LensLike' f s x a

```

This makes two simplifications:

1. Force all lenses to be "monomorphic" by making `s ~ t` and `a ~ b`.
2. Fix these to be real lenses (using `Functor`) rather than generalizing over
`f` to support, say, traversals (which would require `Applicative`).

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/208)

### DIFF
--- a/lens-labels/src/Lens/Labels.hs
+++ b/lens-labels/src/Lens/Labels.hs
@@ -35,7 +35,6 @@ module Lens.Labels (
     Lens,
     Lens',
     -- * HasLens
-    HasLens(..),
     Proxy#,
     proxy#,
     HasLens'(..),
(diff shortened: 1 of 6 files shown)


judah:
Examples of the new improved error messages: https://gist.github.com/judah/e6678257cc28f2c486a631e49db885f2
(That doc was originally written to #118, the last time I took a stab at improving things.)

PR #210, opened 2018-08-23
Don't use `data-default-class` to build protos.

### LINKED ISSUE #194: Define our own "def" rather than the one from data-default-class
The `data-default-class` package is...somewhat controversial.  The main criticism is that `def` doesn't satisfy any laws, so tends to be abused.  See for example the following haskell-cafe thread: https://mail.haskell.org/pipermail/haskell-cafe/2018-May/129053.html

Indeed, from our experience using `proto-lens` on a larger team, we've found that `data-default` creeps into places it's not really intended for; e.g., using it for an integer value instead of an explicit `0`.

However, for protobufs we *do* have a useful law, namely, `encodeMessage def == ""`.

Note that the `proto-lens` library already reexports `def` directly: (`Data.ProtoLens.Message` and `Data.ProtoLens`)

I propose the following changes:

- Remove all our dependencies on `data-default-class` (and `data-default`)
- Change the `Message` class to define its own method `def`:

      class Message m where
          ...
          def :: m

Note that this is a breaking change, so it would cause a major version bump (i.e., `proto-lens-0.4`).  Given that we just released `0.3` which has a fair bit of API churn, I'd probably rather wait a bit before doing another round.  However, it should be fairly easy to write user code that's compatible with both versions, by writing `import Data.ProtoLens (def)`.

judah:
Fixes #194.  `data-default` is problematic since it's (1) lawless and (2)
creeps into places it's not really intended for; e.g., using it for an integer
value instead of an explicit 0.  But protobufs *do* have a useful law, namely,
`encodeMessage def == ""`.

Before this change, the `Message` class had:

```
class Data.Default.Class.Default a => Message a where
    ...
```

After this change:

```
class Message a where
    defMessage :: a
    ...
```

We also provide a module `Data.ProtoLens.Default` which exports `def = defMessage`,
to make it easier to convert old code.

Also bumps versions; in particular, `proto-lens` is now at `v0.4`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/210)


proto-lens/src/Data/ProtoLens/Message.hs
```diff
@@ -71,12 +69,19 @@ import Data.ProtoLens.Encoding.Wire
...
-class Default msg => Message msg where
+class Message msg where
     -- | A unique identifier for this type, of the format
     -- @"packagename.messagename"@.
     messageName :: Proxy msg -> T.Text
+
+    -- | A message with all fields set to their default values.
+    --
+    -- Satisfies @encodeMessage defMessage == ""@ and @decodeMessage "" == Right defMessage@.
+    defMessage :: msg
```
blackgnezdo: Is it too late to rename this to 'defM' or 'defMsg'?

judah:
[defMessage PR](https://trello.com/c/PeLNtpg3/6-defmessage-pr)

PR #215, opened 2018-08-29
Fix a bug in CI that ignored the result of `stack test`.

judah:
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/215)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ script:
       stack)
         STACK_ARGS=(--haddock --no-haddock-deps --ghc-options="-Wall -Werror");
         $STACK build --bench --no-run-benchmarks "${STACK_ARGS[@]}" && $STACK test "${STACK_ARGS[@]}" &&
-        cd proto-lens-tutorial && $STACK build "${STACK_ARGS[@]}" && cd ..;;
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -62,7 +62,7 @@ script:
   - case "$BUILD" in
       stack)
         STACK_ARGS=(--haddock --no-haddock-deps --ghc-options="-Wall -Werror");
-        $STACK build --bench --no-run-benchmarks "${STACK_ARGS[@]}" && $STACK test "${STACK_ARGS[@]}";
+        $STACK build --bench --no-run-benchmarks "${STACK_ARGS[@]}" && $STACK test "${STACK_ARGS[@]}" &&
         cd proto-lens-tutorial && $STACK build "${STACK_ARGS[@]}" && cd ..;;
```
blackgnezdo: What's the significance of the last cd ..? Does this syntax support subshells with ()?

SIde questions: what will people stop making things that looks like shell scripts without being ones? (Is make the first known offender?)

judah: Yeah, pushed a change to use `()`.  It's probably technically not needed anymore, but probably better to be hygenic in case this needs to be changed further or another step added.

As far as I can tell from looking at the output, travis pipes the contents of each entry (everything after the `-`) into a shell, so it should accept the same shell syntax.

Issue #199, opened 2018-05-23
Cannot handle enum variants that begin with non-constructor identifiers?


thoughtpolice:
I'd like to use `proto-lens-protoc` in order to generate Haskell code for working with [ONNX](https://onnx.ai/) protocol buffer images. However, there's a problem: the `enum Version` field of `onnx.proto3`, [as implemented here](https://github.com/onnx/onnx/blob/master/onnx/onnx-ml.proto3#L62), starts one of its enum variants with an underscore:

```protobuf
enum Version {
  // proto3 requires the first enum value to be zero.
  // We add this just to appease the compiler.
  _START_VERSION = 0;
  ...
```

This basically makes it impossible to use `onnx` with `lens-protoc`, because protoc generates Haskell code that looks like

```hs
data Version = _START_VERSION | ...
```

but `_` is not a valid character for beginning a constructor.

---

As far as I know, there's no clear way to work around this other than doing something like committing the files to a repository and modifying them yourself, say `sed -i"" s/_START_VERSION/START_VERSION/g` or whatever on every Haskell file, but this is obviously pretty awful!

Would it be possible for example to add `Setup.hs` utilities that allow you to control some of the generation options, maybe like lens does with it's Template Haskell helpers? Ideally I'd like some predicate like:

```hs
-- possibly unmangle a generated token name by stripping
-- `_` characters from the front, if they exist
unmangle ('_':xs) = xs
unmangle xs = xs

main = defaultMainGeneratingProtosWithNameMangler "..." unmangle
```

---

Alternatively, maybe you could always just strip some characters from the start of `enum` variants if they begin with "bad" characters like underscore. Of course, then `_START_VERSION` and `START_VERSION` would conflict. You could detect this and just say this is a terrible idea and have the compiler fail, or ignore it (and let GHC fail later upon Haskell compilation, when it sees conflicting identifiers), but this seems pretty unlikely to occur...

blackgnezdo:
@thoughtpolice, I think either workaround you proposed (automatic escaping or user supplied mangler) would work. Feel free to send a PR, I'm not sure if anybody will get to this faster than the person whose itch needs scratching.

judah:
I would vote for just modifying the existing auto-name-mangling rules, rather than supplying a user-specific override.

Two other possibilities I can think of:
1. Behave like the `go` binding, and solve it by replacing the initial `_` with an `X` (for example: `_START_VERSION` becomes `XStartVersion`):
[...]

PR #238, opened 2018-09-28
Change enum generation to prevent invalid names

Taneb:
An attempt to fix https://github.com/google/proto-lens/issues/199

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/238)


judah:
Looks good, thank you for the fix!  Can you please add a unit test for this behavior?  The natural place is probably `proto-lens-tests/tests/names{_test.hs,.proto}`.

PR #243, opened 2018-10-09
Improve the bootstrap script.

judah:
- Don't bother calling `clean`; the latest Stack seems better at managing
  this setup.
- Confirm at the end that the generated code builds successfully.


judah:
It's here: https://github.com/google/proto-lens#building-from-head
(Version 1.7.1 is the one I was using to test out this change.)

PR #265, opened 2018-11-29
Use TypeError to improve type error messages.

judah:
Adds an overlapping instance of `HasField` which gives
a better error message when no instance can be found.

Compare these results to:
https://gist.github.com/judah/e6678257cc28f2c486a631e49db885f2#file-lens-errors-md

New error message for the expression
`def & value .~ 3 :: DescriptorProto` :

```
<interactive>:6:14: error:
    • Type DescriptorProto has no field named "value"
    • In the first argument of ‘(.~)’, namely ‘#value’
      In the second argument of ‘(&)’, namely ‘#value .~ "hello"’
      In the expression:
          defMessage & #value .~ "hello" :: DescriptorProto
```

New error message for the expression `() & #name .~ "hello"`:

```
<interactive>:6:6: error:
    • Type () has no field named "name"
    • In the first argument of ‘(.~)’, namely ‘#name’
      In the second argument of ‘(&)’, namely ‘#name .~ "hello"’
      In the expression: () & #name .~ "hello"
Previous error message:
```

In order for this improvement, needed to change the HasField' class slightly,
since overlapping instances don't work with fundeps (or, similarly, open type
families).

For example, previously we had:

```
class HasField' s (x :: Symbol) a | s x -> a where ...

instance HasField' Foo "fieldA" Int32 where ...
instance HasField' Foo "fieldB" Text where ...
```

In this change, we remove the fundep and move that constraint into
each instance:

```
class HasField' s (x :: Symbol) a where ...

instance a ~ Int32 => HasField' Foo "fieldA" where ...
instance a ~ Text => HasField' Foo "fieldB" a where ...
```

I don't think this is a big burden since all those instances are
generated automatically, and we also put in the Haddocks a more
human-readable list of fields for each type.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/265)

### DIFF
--- a/lens-labels/src/Lens/Labels.hs
+++ b/lens-labels/src/Lens/Labels.hs
@@ -86,10 +86,13 @@ instance
 -- Note: in order to support better type error messages, this class does not have
 -- a fundep relationship to the parameter @a@.  Instead, instances are expected to
 -- be defined such that @a@ is derivable from @s@ and @x@.  For example, to define a
--- field named @"x"@ of type @Int32@ for the containing type @Foo@:
+-- field named @"r"@ of type @Int32@ for the containing type @Foo@, use:
 --
--- > instance a ~ Int32 => HasLens' Foo "x" a where ...
+-- > instance a ~ Int32 => HasLens' Foo "r" a where ...
 --
+-- The type checker can use such an instance as long as it knows the types of the
+-- `s` and `x` parameters.  In contrast, an @instance HasLens' Foo "r" Int32@ would
+-- not be used unless the type of `a` is also already known to be `Int32`.
 class HasLens' s (x :: Symbol) a where
     lensOf' :: Functor f => Proxy# x -> (a -> f a) -> s -> f s
 
@@ -101,7 +104,7 @@ type MissingInstanceError s (x :: Symbol) =
 type ASetter s t a b = LensLike Identity s t a b
 
 (.~), set :: ASetter s t a b -> b -> s -> t
-f .~ x' = f %~ const x'
+f .~ x = f %~ const x
 set = (.~)
 
 infixr 4 .~
--- a/lens-labels/src/Lens/Labels.hs
+++ b/lens-labels/src/Lens/Labels.hs
@@ -14,13 +14,13 @@ TODO: support more general optic types (e.g., prisms).
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE ExplicitNamespaces #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 #if __GLASGOW_HASKELL__ >= 802
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -55,7 +55,7 @@ module Lens.Labels (
 import qualified Control.Category as Category
 import GHC.Prim (Proxy#, proxy#)
 import GHC.OverloadedLabels (IsLabel(..))
-import GHC.TypeLits (Symbol)
+import GHC.TypeLits (Symbol, TypeError, ErrorMessage(..))
 
 import Data.Function ((&))
 import Data.Functor.Const (Const(..))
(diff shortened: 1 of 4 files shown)


blackgnezdo:
> I'm not sure why CI is failing for the Cabal builds; maybe something changed in the PPA upstream? It seems unrelated to this change, though, since all the Stack builds are passing.

Agreed, the error is indicative of some infrastructure problem. My guess would be that apt-get suddenly became stricter and no longer allows sha1:
```
Reading package lists...
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install cabal-install-2.0 ghc-8.2.1" failed and exited with 100 during .
```

lens-labels/src/Lens/Labels.hs
```diff
@@ -82,13 +82,26 @@ instance
...
 
 -- | A type class for lens fields.
-class HasLens' s (x :: Symbol) a | s x -> a where
+--
+-- Note: in order to support better type error messages, this class does not have
+-- a fundep relationship to the parameter @a@.  Instead, instances are expected to
+-- be defined such that @a@ is derivable from @s@ and @x@.  For example, to define a
+-- field named @"x"@ of type @Int32@ for the containing type @Foo@:
+--
+-- > instance a ~ Int32 => HasLens' Foo "x" a where ...
```
blackgnezdo: Could you comment on why this is preferrable
`instance a ~ Int32 => HasLens' Foo "x" a where ...`
to this:
`instance HasLens' Foo "x" Int32 where ...`

lens-labels/src/Lens/Labels.hs
```diff
@@ -82,13 +82,26 @@ instance
...
+instance {-# OVERLAPPABLE #-} TypeError (MissingInstanceError s x) => HasLens' s x a where
+    lensOf' = error "Missing HasLens' instance"
+
+type MissingInstanceError s (x :: Symbol) =
+    'Text "Type " ':<>: 'ShowType s ':<>: 'Text " has no field named " ':<>: 'ShowType x
 type ASetter s t a b = LensLike Identity s t a b
 
 (.~), set :: ASetter s t a b -> b -> s -> t
-f .~ x = f %~ const x
+f .~ x' = f %~ const x'
```
blackgnezdo: Why is this prime required? Did we magic up an `x` in global scope somehow?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -284,13 +284,13 @@ generateMessageDecls fieldModName syntaxType env protoName info =
       $ deriving' ["Prelude.Show", "Prelude.Eq", "Prelude.Ord"]
     | oneofInfo <- messageOneofFields info
     ] ++
-    -- instance HasLens' Foo "foo" Bar
+    -- instance Bar ~ a => HasLens' Foo "foo" a where
     --   lensOf _ = ...
     -- Note: for optional fields, this generates an instance both for "foo" and
     -- for "maybe'foo" (see plainRecordField below).
-    [ uncommented $ instDecl []
+    [ uncommented $ instDecl [equalP "a" (tyParen t)]
```
blackgnezdo: I think this results in `a ~ Bar` rather than `Bar ~ a` in the comment above. Make them consistent?

lens-labels/src/Lens/Labels.hs
blackgnezdo: I think I recognized the constraint trick, still makes sense to document it as such.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
judah: Good catch.  Switched the comment.

lens-labels/src/Lens/Labels.hs
judah: Removed.  (It was part of debugging that I forgot to remove.)

lens-labels/src/Lens/Labels.hs
judah: Added more explanation.

lens-labels/src/Lens/Labels.hs
```diff
@@ -86,10 +86,13 @@ instance
 -- Note: in order to support better type error messages, this class does not have
 -- a fundep relationship to the parameter @a@.  Instead, instances are expected to
 -- be defined such that @a@ is derivable from @s@ and @x@.  For example, to define a
--- field named @"x"@ of type @Int32@ for the containing type @Foo@:
+-- field named @"r"@ of type @Int32@ for the containing type @Foo@, use:
```
blackgnezdo: Might as well give it a longer name to reduce the chance of collision?

PR #272, opened 2018-12-10
Bump the lowest supported LTS to lts-8 (ghc-8.0.2).

judah:
Lets us remove support for lts-8 (ghc-8.0.1), including some
CPP directives for older haskell-src-exts and optparse-applicative.

For bookkeeping I increased the lower bound on optparse-applicative
in proto-lens-optparse, even though technically we didn't lose
support for older versions.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/272)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,10 @@ cache:
 
 matrix:
   include:
-  - env: BUILD=cabal GHCVER=8.0.1 CABALVER=1.24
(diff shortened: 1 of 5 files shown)


judah:
For reference: ghc-8.0.2 was released on 2017-1-11, and lts-8.0 was released on 2018-2-12.

PR #273, opened 2018-12-12
Expose Data.ProtoLens.Encoding.Bytes.

judah:
It'll be used by the generated encodings.

Also add `runParser` and `runBuilder` functions, and use them
to simplify `Data.ProtoLens.Encoding.Reflected` a little.

### DIFF
--- a/proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
+++ b/proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
@@ -10,13 +10,21 @@
 
 -- | Utility functions for parsing and encoding individual types.
 module Data.ProtoLens.Encoding.Bytes(
+    -- * Running encodings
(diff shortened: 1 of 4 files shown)


proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
```diff
@@ -34,6 +44,12 @@ import Foreign.Marshal.Alloc (alloca)
 import Foreign.Storable (Storable, peek, poke)
 import System.IO.Unsafe (unsafePerformIO)
 
+runParser :: Parser a -> ByteString -> Either String a
```
blackgnezdo: Maybe add a bit of haddock for public functions?

judah: Done.

PR #281, opened 2018-12-13
Make the decisions about map fields in collectDefinitions.

judah:
Previously Generate* had some of the responsibility, which
complicated the code (e.g., requiring passing `Env` to more
functions).

For efficiency, when looking for a map entry type we only look
in the immediate children (rather than transitively to all
nested types).  This required using a Tree as an intermediate
datastructure, but it didn't seem to complicate the code much.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/281)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -183,6 +188,7 @@ data FieldKind
         -- Stored internally as either a list or a map, depending on
         -- whether the field's FieldDescriptorProto of the field has
         -- options.map_entry set.
+    | MapField MapEntryInfo
```
blackgnezdo: Haddock?

judah:
Closing since it was merged as part of #283.

PR #283, opened 2018-12-13
Generated groups

judah:
Support groups in the generated encodings.
    
I changed the `Definitions` module to keep track of the field number
of each group message type.  It meant a little more bookkeeping, but
simpified the `Generate.Encodings` module.  (Similar to #281, it
helps avoid the need to access an `Env` when generating the encodings.)

This PR depends on #281.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/283)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -209,6 +217,18 @@ data OneofCase = OneofCase
         -- ^ The name for building 'Prism' definition.
     }
 
+-- | The "entry" type for a map field is an proto message, autogenerated by
```
blackgnezdo: is a?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -355,28 +380,75 @@ messageDefs syntaxType protoPrefix hsPrefix d
...
+                    (d ^. enumType)
+    -- For efficiency, only look for map entries within the immediate
+    -- nested types, rather than recursively searching through all of them.
+    mapEntries = collectMapEntries $ map rootLabel subDefs
+
+-- | If this type is a map entry, retrieves the relevant information
+-- along with the proto name of this type.
+mapEntryInfo :: Definition Name -> Maybe MapEntryInfo
+mapEntryInfo (Message m)
+    | messageDescriptor m ^. options . mapEntry
```
blackgnezdo: I don't understand this construct. Could you rephrase this as a simple case?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -355,28 +380,75 @@ messageDefs syntaxType protoPrefix hsPrefix d
...
+-- | Returns a map whose keys are the proto type name of the entry message.
+collectMapEntries :: [(Text, Definition Name)] -> Map.Map Text MapEntryInfo
+collectMapEntries defs =
+    Map.fromList
+        [(protoName, e) | (protoName, d) <- defs, Just e <- [mapEntryInfo d]]
+
+-- | A map whose keys are the proto type names of the groups in a message,
+-- and values are the field numbers for those groups.
+-- (Every group corresponds to exactly one message field.)
+type GroupMap = Map.Map Text Int32
```
blackgnezdo: Idle musing, I wonder if Text for such things is justified or if we would be better off if all this utf8 stuff remained as ByteString of some sort. It's likely impossible to measure any difference between the two.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
```diff
@@ -355,28 +380,75 @@ messageDefs syntaxType protoPrefix hsPrefix d
...
+
+collectGroupFields :: [FieldDescriptorProto] -> Map.Map Text Int32
```
blackgnezdo: -> GroupMap?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
blackgnezdo: Ah, this is a boolean guard, no need to do anything here.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
judah: Good catch, done.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
judah: Done

proto-lens-protoc/src/Data/ProtoLens/Compiler/Definitions.hs
judah: I feel the ability to distinguish between utf8 and raw bytes in the types of the APIs we provide is pretty important.  So I'd want to see a measurable performance issue before strongly considering something else.

There are a handful of Haskell packages that support utf-8-backed Text, e.g `text-short`, but I'm not sure any of them is really in wide use yet.

PR #288, opened 2018-12-17
Check for required fields in the generated parser.

judah:
For each required field, adds a new `Bool` argument to the main loop.
When the loop is finished, it checks that all of the fields are set.

Also adds a unit test checking that we report *all* names of missing
required fields, not just the first one that we find.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/288)

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
@@ -1,6 +1,7 @@
 -- | This module generates code for decoding and encoding protocol buffer messages.
 --
 -- Upstream docs: https://developers.google.com/protocol-buffers/docs/encoding
+{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE LambdaCase #-}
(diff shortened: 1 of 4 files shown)


proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
```diff
@@ -70,13 +86,78 @@ generatedParser m =
...
+        , let n = overloadedName (fieldName $ plainFieldInfo f)
+        , RequiredField <- [plainFieldKind f]
+        ]
+    }
+
+-- | The intial values of the loop arguments.
+initialParseState :: ParseState a -> ParseState Exp
+initialParseState s = ParseState
+    { partialMessage = "Data.ProtoLens.defMessage"
+    , requiredFieldsUnset = fmap (const "Prelude.True")
```
blackgnezdo: A rare case were <$> looks like a very nice win.

judah: Done.

PR #293, opened 2018-12-18
Remove the Encoding.Reflected API.

### LINKED ISSUE #125: Don't fail to decode if fields have the wrong wire type
If a field has the wrong wire type, currently we fail the decode.  Instead, we should just ignore that field (and/or add it to the unknown fields set, once we've implemented #29).

I confirmed that C++, Java and Go all have this more lenient behavior (although it's not documented well).

Also mentioned here:
https://github.com/apple/swift-protobuf/issues/342

judah:
Also renames `unfinished{Parse,Build}Message` to `{parse,build}Message`.

NOTE: this fixes #125 as a result of switching to the new implementation.  If
fields have the wrong wire type, we now store them in `unknownFields` rather
than failing the parse.  I updated a few unit tests that expected the old
behavior.

The only other unit test change that I made was to text_format_test, which
had a test that relied on the order in which fields are serialized to
the wire.  Previously, it was in order of tag type; now, it is (implicitly)
based on the order in the `.proto` file format.

There's some cleanup still to be done around the *.Wire modules,
but I'll do that as a follow-up.

Also re-bootstraps to get the latest

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/293)


proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
```diff
@@ -1,29 +1,112 @@
...
--- | Parse an unknown WireValue, given that the type+number has
--- already been parsed.
-parseTaggedValue :: Word64 -> Parser TaggedValue
-parseTaggedValue w = do
-    (SomeWireType wt, tag) <- runEither $ splitTypeAndTag w
-    value <- getWireValue wt
-    return $ TaggedValue tag (WireValue wt value)
+-- | A tag that identifies a particular field of the message when converting
+-- to/from the wire format.
+newtype Tag = Tag { unTag :: Int}
```
blackgnezdo: Regularize spaces?

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
```diff
@@ -1,29 +1,112 @@
...
+buildWireValue (Fixed64 w) = putFixed64 w
+buildWireValue (Fixed32 w) = putFixed32 w
+buildWireValue (Lengthy b) =
+    putVarInt (fromIntegral $ B.length b)
+    <> putBytes b
+buildWireValue StartGroup = mempty
+buildWireValue EndGroup = mempty
+
+wireValueToInt :: WireValue -> Word8
+wireValueToInt VarInt{} = 0
```
blackgnezdo: If this encoding corresponds to anything external, link to the spec?

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
```diff
@@ -1,29 +1,112 @@
...
+splitTypeAndTag :: Word64 -> (Tag, Word8)
+splitTypeAndTag w = (fromIntegral $ w `shiftR` 3, fromIntegral (w .&. 7))
+
+joinTypeAndTag :: Tag -> Word8 -> Word64
+joinTypeAndTag (Tag t) w = fromIntegral t `shiftL` 3 .|. fromIntegral w
+
+parseFieldSet :: Parser FieldSet
+parseFieldSet = loop []
+  where
+    loop !ws = do
```
blackgnezdo: Does ! do anything meaningful here?

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding/Wire.hs
judah: Removed.

Also, after taking another look, I added a `$!` to force the result of `reverse` rather than returning a thunk.

PR #302, opened 2019-01-07
Store repeated fields as Vectors.

### LINKED ISSUE #58: Please use vectors instead of lists for packed values

judah:
Fixes #58.

For every repeated field `foo`, add a new lens `vector'foo`
whose field type is a `Vector` instead of a list.  Internally,
repeated fields are always stored as `Vectors`, whether or not
the field is actually packed.  Specifically:
- `Data.Vector.Unboxed.Vector` if the field's Haskell representation is
  fixed-length (`Int32`, `Float`, etc.), or
- `Data.Vector.Vector` (i.e., boxed) otherwise (`Text`, `ByteString`, messages,
  etc).

For backwards compatibility, the name `foo` continues to be used
for a list and is implemented as

    foo = vector'foo . iso Vector.toList Vector.fromList

The generated encoding/decoding code uses those `vector'*` lenses
internally.  For parsing, we keep a separate mutable vector for each
repeated field and grow it manually using the
`Data.ProtoLens.Encoding.Growing` module.

Two potentially dodgy things about this change:
1. I couldn't come up with an efficient, simple, and fully safe API
   for interleaving the vector mutation and parsing effects.  I settled
   for exposing `unsafeLiftIO :: IO a -> Parser a` in a separate module
   `Data.ProtoLens.Encoding.Parser.Unsafe`.
   The standard ST functions (e.g. `create` and `createT` from `vector`)
   don't really work here because we have an arbitary set of vectors
   of different element types.
2. In order to get the inner loops to optimize well, I introduced
   a function (born of much testing and staring at Core output):
   ```Data.ProtoLens.Encoding.Bytes.foldMapBuilder :: Vector v a => (a -> Builder) -> v a -> Builder```
   It uses internal Builder functions but produces large speedups
   due to getting GHC to inline the inner loop correctly.  (In fact,
   without it this change is slightly *worse* in encoding performance
   than the version with lists.)

TODO:
- Store enums in unboxed vectors.
- Also store unknown fields in a Vector (boxed) rather than a list.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/302)

### DIFF
--- a/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
+++ b/proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
@@ -46,6 +46,10 @@ import Proto.Google.Protobuf.Descriptor_Fields
 import Data.ProtoLens.Compiler.Combinators
 import Data.ProtoLens.Compiler.Definitions
 import Data.ProtoLens.Compiler.Generate.Encoding
+import Data.ProtoLens.Compiler.Generate.Field
+    ( hsFieldType
+    , hsFieldVectorType
+    )
 
 -- Whether to import the "Runtime" modules or the originals;
 -- e.g., Data.ProtoLens.Runtime.Data.Map vs Data.Map.
@@ -97,12 +101,17 @@ generateModule modName imports modifyImport definitions importedEnv services
               [ "Prelude", "Data.Int", "Data.Monoid", "Data.Word"
               , "Data.ProtoLens"
               , "Data.ProtoLens.Encoding.Bytes"
+              , "Data.ProtoLens.Encoding.Growing"
+              , "Data.ProtoLens.Encoding.Parser.Unsafe"
               , "Data.ProtoLens.Encoding.Wire"
               , "Data.ProtoLens.Message.Enum"
               , "Data.ProtoLens.Service.Types"
               , "Lens.Family2", "Lens.Family2.Unchecked"
               , "Data.Text",  "Data.Map", "Data.ByteString", "Data.ByteString.Char8"
               , "Data.Text.Encoding"
+              , "Data.Vector"
+              , "Data.Vector.Generic"
+              , "Data.Vector.Unboxed"
               , "Lens.Labels", "Text.Read"
               ]
             ++ map importSimple imports
@@ -262,7 +271,7 @@ generateMessageDecls fieldModName env protoName info =
     -- haskell: data Foo'Bar = Foo'Bar'c !Prelude.Float
     --                       | Foo'Bar's !Sub
     [ uncommented $ dataDecl (oneofTypeName oneofInfo)
-      [ conDecl consName [hsFieldType env $ fieldDescriptor f]
+      [ conDecl consName [hsFieldType env f]
       | c <- oneofCases oneofInfo
       , let f = caseField c
       , let consName = caseConstructorName c
@@ -340,7 +349,7 @@ generatePrisms env oneofInfo =
                                 -- The oneof sum type name
                              @@ (tyCon . unQual $ oneofTypeName oneofInfo)
                                 -- The field contained in the sum
-                             @@ (hsFieldType env $ fieldDescriptor f)
+                             @@ (hsFieldType env f)
         -- Generate function definition
         -- Prism' is constructed with Constructor for building value
         -- and Deconstructor and wrapping in Just for getting value
@@ -672,8 +681,8 @@ plainRecordField env (PlainFieldInfo kind f) = case kind of
         -- type instance Field "foo" Foo = Map Bar Baz
     MapField entry ->
             let mapType = "Data.Map.Map"
-                            @@ hsFieldType env (fieldDescriptor $ keyField entry)
-                            @@ hsFieldType env (fieldDescriptor $ valueField entry)
+                            @@ hsFieldType env (keyField entry)
+                            @@ hsFieldType env (valueField entry)
             in recordField mapType
                   [LensInstance
                        { lensSymbol = baseName
@@ -683,23 +692,36 @@ plainRecordField env (PlainFieldInfo kind f) = case kind of
         -- data Foo = Foo { _Foo_bar :: [Bar] }
         -- type instance Field "bar" Foo = [Bar]
     RepeatedField {} ->
-            recordField listType
-                  [LensInstance
+            recordField vectorType
+                  [ LensInstance
                       { lensSymbol = baseName
                       , lensFieldType = listType
+                      , lensExp = vectorAccessor
+                      }
+                  , LensInstance
+                      { lensSymbol = "vector'" <> baseName
+                      , lensFieldType = vectorType
                       , lensExp = rawAccessor
-                      }]
+                      }
+                  ]
   where
     recordField = RecordField (haskellRecordFieldName $ fieldName f)
     baseName = overloadedName $ fieldName f
     fd = fieldDescriptor f
-    baseType = hsFieldType env fd
+    baseType = hsFieldType env f
     maybeType = "Prelude.Maybe" @@ baseType
     listType = tyList baseType
+    vectorType = hsFieldVectorType f @@ baseType
     rawAccessor = "Prelude.id"
     maybeAccessor = "Data.ProtoLens.maybeLens"
                           @@ hsFieldValueDefault env fd
 
+vectorAccessor :: Exp
+vectorAccessor = "Lens.Family2.Unchecked.lens" @@ getter @@ setter
+  where
+    getter = "Data.Vector.Generic.toList"
+    setter = lambda ["_", "y__"]
+                $ "Data.Vector.Generic.fromList" @@ "y__"
 
 oneofRecordField :: Env QName -> OneofInfo -> RecordField
 oneofRecordField env oneofInfo
@@ -746,39 +768,9 @@ oneofRecordField env oneofInfo
             | c <- oneofCases oneofInfo
             , let f = caseField c
             , let baseName = overloadedName $ fieldName f
-            , let baseType = hsFieldType env $ fieldDescriptor f
+            , let baseType = hsFieldType env f
             , let maybeName = "maybe'" <> baseName
             ]
-
-hsFieldType :: Env QName -> FieldDescriptorProto -> Type
-hsFieldType env fd = case fd ^. type' of
-    FieldDescriptorProto'TYPE_DOUBLE -> "Prelude.Double"
-    FieldDescriptorProto'TYPE_FLOAT -> "Prelude.Float"
-    FieldDescriptorProto'TYPE_INT64 -> "Data.Int.Int64"
-    FieldDescriptorProto'TYPE_UINT64 -> "Data.Word.Word64"
-    FieldDescriptorProto'TYPE_INT32 -> "Data.Int.Int32"
-    FieldDescriptorProto'TYPE_FIXED64 -> "Data.Word.Word64"
-    FieldDescriptorProto'TYPE_FIXED32 -> "Data.Word.Word32"
-    FieldDescriptorProto'TYPE_BOOL -> "Prelude.Bool"
-    FieldDescriptorProto'TYPE_STRING -> "Data.Text.Text"
-    FieldDescriptorProto'TYPE_GROUP
-        | Message m <- definedFieldType fd env -> tyCon $ messageName m
-        | otherwise -> error $ "expected TYPE_GROUP for type name"
-                              ++ unpack (fd ^. typeName)
-    FieldDescriptorProto'TYPE_MESSAGE
-        | Message m <- definedFieldType fd env -> tyCon $ messageName m
-        | otherwise -> error $ "expected TYPE_MESSAGE for type name"
-                              ++ unpack (fd ^. typeName)
-    FieldDescriptorProto'TYPE_BYTES -> "Data.ByteString.ByteString"
-    FieldDescriptorProto'TYPE_UINT32 -> "Data.Word.Word32"
-    FieldDescriptorProto'TYPE_ENUM
-        | Enum e <- definedFieldType fd env -> tyCon $ enumName e
-        | otherwise -> error $ "expected TYPE_ENUM for type name"
-                              ++ unpack (fd ^. typeName)
-    FieldDescriptorProto'TYPE_SFIXED32 -> "Data.Int.Int32"
-    FieldDescriptorProto'TYPE_SFIXED64 -> "Data.Int.Int64"
-    FieldDescriptorProto'TYPE_SINT32 -> "Data.Int.Int32"
-    FieldDescriptorProto'TYPE_SINT64 -> "Data.Int.Int64"
 
 hsFieldDefault :: Env QName -> PlainFieldInfo -> Exp
 hsFieldDefault env f = case plainFieldKind f of
@@ -786,7 +778,7 @@ hsFieldDefault env f = case plainFieldKind f of
     OptionalValueField -> hsFieldValueDefault env fd
     OptionalMaybeField -> "Prelude.Nothing"
     MapField {} -> "Data.Map.empty"
-    RepeatedField {} -> list []
+    RepeatedField {} -> "Data.Vector.Generic.empty"
   where
     fd = fieldDescriptor (plainFieldInfo f)
 
@@ -901,7 +893,7 @@ messageInstance env protoName m =
                   [ fieldUpdate (unQual $ messageUnknownFields m)
                         "[]"]
       ]
-    , [ match "parseMessage" [] $ generatedParser m ]
+    , [ match "parseMessage" [] $ generatedParser env m ]
     , [ match "buildMessage" [] $ generatedBuilder m ]
     ]
   where
@@ -950,7 +942,7 @@ fieldDescriptorExpr env n f =
         @@ (fieldTypeDescriptorExpr (fd ^. type')
                 @::@
                     ("Data.ProtoLens.FieldTypeDescriptor"
-                        @@ hsFieldType env fd))
+                        @@ hsFieldType env (plainFieldInfo f)))
         @@ fieldAccessorExpr f)
     -- TODO: why is this type sig needed?
     @::@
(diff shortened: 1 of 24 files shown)


judah:
Benchmark results: https://docs.google.com/spreadsheets/d/1XSmrbGGsE3ntlv__ytX7obPtjMY_CHc7FkAWdsUNw68/edit?usp=sharing

blackgnezdo:
> Benchmark results: https://docs.google.com/spreadsheets/d/1XSmrbGGsE3ntlv__ytX7obPtjMY_CHc7FkAWdsUNw68/edit?usp=sharing

Which GHC version/platform for this on?

judah:
The benchmarks were on my MacBook Pro, with `lts-12.7` (so, `ghc-8.4.3`).

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
```diff
@@ -74,28 +77,66 @@ generatedParser m =
...
+    | (frozen, mutable) <- Map.elems $ Map.intersectionWith (,)
+                                frozenNames (repeatedFieldMVectors s)
+    ]
+    ++
+    [ stmt $ checkMissingFields s
+    , stmt $ "Prelude.return" @@
+        (over' unknownFields' "Prelude.reverse"
+            @@(foldr (\(finfo, frozen) x' -> "Lens.Family2.set"
+                                            @@ lensOfVectorField finfo
+                                            @@ var (unQual frozen) @@ x')
```
blackgnezdo: maybe move @@ x' below to line up all args?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
```diff
@@ -74,28 +77,66 @@ generatedParser m =
...
+                        @@ mutable)
+    | (frozen, mutable) <- Map.elems $ Map.intersectionWith (,)
+                                frozenNames (repeatedFieldMVectors s)
+    ]
+    ++
+    [ stmt $ checkMissingFields s
+    , stmt $ "Prelude.return" @@
+        (over' unknownFields' "Prelude.reverse"
+            @@(foldr (\(finfo, frozen) x' -> "Lens.Family2.set"
+                                            @@ lensOfVectorField finfo
```
blackgnezdo: The 'lens' part doesn't seem to be reflected in the pseudo code above. Am I missing something?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
```diff
@@ -74,28 +77,66 @@ generatedParser m =
...
+                                            @@ var (unQual frozen) @@ x')
+                     (partialMessage s)
+                     (Map.elems $ Map.intersectionWith (,)
```
blackgnezdo: Is partially applying the lambda above instead of (,) any better/worse/pointlessly different?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
```diff
@@ -104,11 +145,13 @@ data ParseState v = ParseState
     , requiredFieldsUnset :: Map.Map FieldId v
         -- ^ The required fields of the message, each corresponding to
         -- a @Bool@ argument of the loop.
+    , repeatedFieldMVectors :: Map.Map FieldId v
```
blackgnezdo: Add haddock?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
```diff
@@ -104,11 +145,13 @@ data ParseState v = ParseState
...
     } deriving Functor
 
 -- | Returns a sequence all arguments of the loop function.
```
blackgnezdo: A sequence _of_ all?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Field.hs
```diff
@@ -6,17 +6,68 @@
...
+    ( hsFieldType
+    , hsFieldVectorType
+    , FieldEncoding(..)
     , fieldEncoding
     , lengthy
     , groupEnd
     , isolatedLengthy
     ) where
 
+import Data.Text (unpack)
```
blackgnezdo: Do you have a particular style you follow for imports sorting?

proto-lens-tests/tests/repeated_test.hs
```diff
@@ -5,24 +5,48 @@
...
 defBar :: Bar
 defBar = defMessage
 
+-- | Test that the vector and list lenses for repeated fields
+-- behave the same.
+vectorTest ::
+    forall a v b . (Eq a, Eq b, Show a, Show b, Message a, V.Vector v b)
+    => String -> Proxy a -> Gen b
+    -> Lens' a [b] -> Lens' a (v b) -> Test
+vectorTest name _ arbitraryElem listLens vecLens = testGroup name
```
blackgnezdo: Do you need a `Proxy a` here or can you get by with \@ a ?

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
```diff
@@ -140,3 +143,24 @@ wordToSignedInt64 n
...
+-- @Builder@s.
+--
+-- This function has been hand-tuned to perform better than a naive
+-- implementation using, e.g., Vector.foldr or a manual loop.
+foldMapBuilder :: V.Vector v a => (a -> Builder) -> v a -> Builder
+foldMapBuilder f = \v0 -> Internal.builder (loop v0)
+    -- Place v0 on the right-hand side so that GHC actually inlines
+    -- this function.
+  where
+    -- Fully-saturate the inner loop (rather than currying away `cont`
```
blackgnezdo: Thanks for adding detailed comments about tuning!

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
```diff
@@ -0,0 +1,59 @@
...
+
+import Control.Monad.Primitive (PrimMonad, PrimState, RealWorld)
+
+import qualified Data.Vector.Generic as V
+import qualified Data.Vector.Generic.Mutable as MV
+
+-- | A mutable vector which can increase in capacity.
+data Growing v s a = Growing
+    {-# UNPACK #-} !Int
+        -- The number of elements in the mutable vector
```
blackgnezdo: Turn into a haddock field comment?

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
```diff
@@ -0,0 +1,59 @@
...
+new = Growing 0 <$> MV.new 0
+
+-- | Unsafely convert a growing vector to an immutable one without
+-- copying.  After this call, you may not use the growing vector
+-- nor any other growing vectors that were used to produce this one.
+unsafeFreeze
+    :: (PrimMonad m, V.Vector v a)
+    => Growing v (PrimState m) a -> m (v a)
+unsafeFreeze (Growing len m)
+    = V.unsafeFreeze (MV.take len m)
```
blackgnezdo: Move up to conserve some vertical space?

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
```diff
@@ -0,0 +1,59 @@
+-- | A mutable vector that grows in size.
```
blackgnezdo: Does it make sense to give a quick code sample of how to use this module safely?

proto-lens/src/Data/ProtoLens/Encoding/Parser/Internal.hs
```diff
@@ -0,0 +1,40 @@
...
+import Foreign.Ptr
+
+-- | A monad for parsing an input buffer.
+newtype Parser a = Parser
+    { unParser :: Ptr Word8 -- End position of the input
+               -> Ptr Word8 -- Current position in the input
+               -> ExceptT String IO (ParserResult a)
+    }
+
+data ParserResult a = ParserResult
```
blackgnezdo: How do you feel about renaming this to `ParserSuccess`? I paused for a second to consider why parser result didn't represent a failure case.

proto-lens/src/Data/ProtoLens/Encoding/Parser/Unsafe.hs
```diff
@@ -0,0 +1,22 @@
...
+-- The generated code uses this function to construct vectors
+-- efficiently by incrementally building up mutable vectors.
+--
+-- NOTE: This is unsafe since @runParser@
+-- is a pure function, which lets us lift arbitrary IO into
+-- pure operations.
+-- However, here are some guarantees that we do get:
+--
+-- - For each individual call to 'runParser', the action
+--   wrapped by unsafeParserLiftIO will be called exactly once.
```
blackgnezdo: Quotes around `unsafeParserLiftIO`?

proto-lens/src/Data/ProtoLens/Encoding/Parser/Unsafe.hs
```diff
@@ -0,0 +1,22 @@
...
+-- - Different calls to 'unsafeLiftIO' within the same call to
+--   'runParser' will be sequenced according to their order in the Parser
+-- monad.
```
blackgnezdo: Is this indentation OK?

judah:
The benchmark results indicate a *significant* further speedup possibility, since decoding packed fixed32s is more than 4x faster than decoding packed floats.  I'll look into that in another PR.

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
judah: It breaks on older versions of Haddock due to the bang pattern.

Since the constructor isn't exposed, it won't show in the generated html so I don't think it matters in practice.

proto-lens/src/Data/ProtoLens/Encoding/Parser/Internal.hs
judah: Makes sense, but I'll do that as a separate PR since it's not really related to this change.

I'll also try benchmarking a change your comment made me realize: we can merge the Either and ParserResult.  Instead of `Either String (ParserResult a)`, have

```
data ParserResult a
   = ParserFailure String
   | ParserSuccess {...}
```

proto-lens/src/Data/ProtoLens/Encoding/Parser/Unsafe.hs
judah: It renders fine, but I fixed it so the source is more legible.

proto-lens/src/Data/ProtoLens/Encoding/Parser/Unsafe.hs
judah: Done.  (Also fixed the name which should be `unsafeLiftIO`.)

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
judah: Done.

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
judah: Done.

proto-lens-tests/tests/repeated_test.hs
judah: Switched to `@`.  Nice!

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Field.hs
judah: Reordered.  Usually I do a sorted block of imports from other packages, followed by a block from the current package.  But I'm sure I haven't been completely consistent.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
judah: Done.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
judah: Done.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
judah: Good catch; I've updated the comment.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
judah: Done

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
```diff
@@ -74,28 +77,66 @@ generatedParser m =
...
+                if' end (finish m exprs)
                     $ do'
                         [ tag <-- getVarInt'
                         , stmt $ case' tag $ parseTagCases continue exprs m
                         ]
             ]
 
+{- | A Parser expression that finalizes the message.
+
+do
```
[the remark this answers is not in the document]
judah: Moved the code block into the body of the function (as we already do in other parts of the code).

judah:
I ran the benchmarks on `lts-13.1` (`ghc-8.4.3`).  The relative results are basically the same.  It did seem like some benchmarks showed a 10-20% hit both before and after this change though.

One caveat: initially, the float conversion in Data.ProtoLens.Encoding.Bytes got slower with 8.6.3 due to a missed inlining.  However, after I incorporated #303 the effect went away (because the relevant code was replaced with GHC primitives) and the performance is the same as with 8.4.3.

judah:
I don't think it's worth benchmarking older compilers.  But I did check the generated core for the manually-written `foldMapBuilder` to confirm that 8.0 and 8.2 treat it the same.

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
```diff
@@ -1,4 +1,16 @@
...
+--
+-- > import qualified Data.ProtoLens.Encoding.Growing as Growing
+-- > import qualified Data.Vector.Unboxed as V
+-- > test :: IO (V.Vector Int)
+-- > test = do
+-- >     v <- Growing.new
+-- >     v' <- Growing.append v 1
+-- >     v'' <- Growing.append v' 2
+-- >     v''' <- Growing.append v'' 3
+-- >     unsafeFreeze v'''
```
blackgnezdo: Come to think of it, does it make sense to turn this into a test? I know all this stuff is well covered by generated code that we test. It might still be useful to have a direct test considering how unsafe the innards are.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -683,23 +692,36 @@ plainRecordField env (PlainFieldInfo kind f) = case kind of
...
-            recordField listType
-                  [LensInstance
+            recordField vectorType
+                  [ LensInstance
                       { lensSymbol = baseName
                       , lensFieldType = listType
+                      , lensExp = vectorAccessor
+                      }
+                  , LensInstance
+                      { lensSymbol = "vector'" <> baseName
```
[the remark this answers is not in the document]
judah: Yeah, bikeshedding seems reasonable.  I'm against a prefix as short as `v`, but `vec` seems pretty reasonable.

I think we should also consider a migration path for eventually deprecating the list lenses and making the `baseName` lens provide a vector.  We can probably punt on it for now though.

Another option I was thinking about was `baseName <> "'vector"`, though that's less consistent with the `maybe'*` lenses.

I think we should *not* CamelCase `baseName`.  First, because it would be different behavior than the `maybe'*` lenses.  And second, because the `'` already acts as a separator so the upper-casing would be redundant and make it look like a constructor rather than a value.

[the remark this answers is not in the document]
blackgnezdo: The list deprecation could be achieved with some kind of option people stick into their .proto. But's a battle for another day.

proto-lens/src/Data/ProtoLens/Encoding/Growing.hs
judah: Ah yeah, good point.  I intended to add unit tests for this package but forgot.  Will do.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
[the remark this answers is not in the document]
blackgnezdo: Looks like we agreed to `vec`.

proto-lens/tests/growing_test.hs
```diff
@@ -0,0 +1,37 @@
...
+
+testFromList :: [Int] -> Property
+testFromList xs = fromListGrowing xs === V.fromList xs
+
+fromListGrowing :: V.Unbox a => [a] -> V.Vector a
+fromListGrowing xs0 = runST (new >>= fill xs0 >>= unsafeFreeze)
+
+fill :: V.Unbox a => [a] -> Growing V.Vector s a -> ST s (Growing V.Vector s a)
+fill [] v = return v
+fill (x:xs) v = append v x >>= fill xs
```
blackgnezdo: Is this foldM in disguise?

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate.hs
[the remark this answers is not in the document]
judah: Done.

proto-lens-protoc/src/Data/ProtoLens/Compiler/Generate/Encoding.hs
judah: Done.

proto-lens/tests/growing_test.hs
judah: Done.

PR #309, opened 2019-01-31
Merge lens-labels into proto-lens.

judah:
The separate "lens-labels" package hasn't been used outside of
proto-lens.  There's also a confusing distinction between
regular lenses and the newtypes in `Lens.Labels` -- which, again,
aren't widely used.

- `Lens.Labels` -> `Data.ProtoLens.Fields`
    - Defines the `HasField` class.
    - Also defines a new `field` function that works with
      `TypeApplications`.
    - Removes the `LensFn` newtype and related functions.
- `Lens.Labels.Unwrapped` -> `Data.ProtoLens.Labels`
    - Defines the orphan `IsLabel` instance.
- `Lens.Labels.Prism` -> `Data.ProtoLens.Prism`
    - Defines prisms and related functions.

Also rename `HasLens` to `HasField` to track the recent GHC proposal:
(https://github.com/ghc-proposals/ghc-proposals/pull/158

Also bump proto-lens-protobuf-types to version 0.5 since it's
now more dependent on the new version of proto-lens.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/309)


proto-lens/src/Data/ProtoLens/Field.hs
```diff
@@ -0,0 +1,57 @@
+{- | An implementation of overloaded record fields.  This module
+enables different types in the same module to have fields of the
+same name.
+
+To use instances from this class, either:
+
+* Enable the @OverloadedLabels@ extension and @import Data.ProtoLens.Labels ()@,
```
blackgnezdo: Missing "or" at the end?

proto-lens/src/Data/ProtoLens/Field.hs
```diff
@@ -0,0 +1,57 @@
...
+* Use the 'field' function along with the @TypeApplications@ extension.
```
blackgnezdo: If we have examples in the tutorial, it makes sense to link to them.

proto-lens/src/Data/ProtoLens/Field.hs
judah: Done.

docs/tutorial.md
```diff
@@ -193,24 +239,37 @@ myBar = defMessage
...
+main = print $ myBar ^. P.bippy
 ```
 
-The second method is by using the `OverloadedLabels` extension and importing the orphan instance of `IsLabel` for `proto-lens` `LensFn` type, giving us the use of `#` for prefixing our field accessors. To bring this instance into scope we need to also import `Data.ProtoLens.Labels`:
-``` haskell
-{-# LANGUAGE OverloadedLabels #-}
+This approach is less flexible, since it may require manual adjustment when the
+same name is defined in two different `.proto` files, and thus exported by both
+of their `*_Fields` modules.  If that happens, you can resolve the conflict
+by  importing the definition from exactly
```
blackgnezdo: Extra space after `by`.

proto-lens/src/Data/ProtoLens/Field.hs
```diff
@@ -4,8 +4,12 @@ same name.
 
 To use instances from this class, either:
 
-* Enable the @OverloadedLabels@ extension and @import Data.ProtoLens.Labels ()@,
+* Enable the @OverloadedLabels@ extension and
+  @import Data.ProtoLens.Labels ()@; or,
 * Use the 'field' function along with the @TypeApplications@ extension.
+* Import the corresponding names from the autogenerated @*_Fields@ module.
```
blackgnezdo: Looks like we have 3 alternatives now, could you check the typography works out for this list?

docs/tutorial.md
judah: Done.

proto-lens/src/Data/ProtoLens/Field.hs
judah: Done.  The formatting looks reasonable in haddock.  I changed it slightly to make the "or" happen in between the second and third options.

judah:
@thumphries @nhibberd FYI, since if I remember correctly Formation was using OverloadedLabels with proto-lens.  Let us know if there are any concerns on your end.

PR #310, opened 2019-02-01
Revert "Use TypeError to improve type error messages. (#265)"

judah:
This reverts commit 60b7c3cd8b8dac9e6138b3c2324df9d8e49629ca.

It turned out that the TypeError would sometimes fire incorrectly,
even when there was transitive instance available.  Details:
https://github.com/google/proto-lens/commit/4bb77a571dbccb43b657e6e52ace34eaa422a705

I suspect a GHC bug, but wasn't easily able to simplify the example enough to know
for sure.

---
This change is [Reviewable](https://reviewable.io/reviews/google/proto-lens/310)

### DIFF
--- a/proto-lens-protoc/Changelog.md
+++ b/proto-lens-protoc/Changelog.md
@@ -5,7 +5,6 @@
 ### Breaking Changes
 - Capitalize enum values, and capitalize names of enum sub-messages (#270).
 - Fix the parser to fail on end-group markers with the wrong tag number (#282).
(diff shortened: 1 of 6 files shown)


blackgnezdo:
If we make a release with this reverted API, will it be a breaking change if we decide to reapply the patch?

judah:
Yes, it would be a breaking change to the API.  (Plus, it's a change that needs to be made to proto-lens and proto-lens-protoc simultaneously.)

PR #317, opened 2019-04-12
Fix a naming conflict between message constructors and enum values.

judah:
In general, protoc doesn't allow message types and enum values to be the same.
However, since we change the case of enum values, it's possible for us
to produce an overlap.  (For example, a message called "Job" and an enum value
called "job").

I fixed it by changing the name of the message constructor, which is an internal detail
not exposed by our API.

### DIFF
--- a/proto-lens-protoc/Changelog.md
+++ b/proto-lens-protoc/Changelog.md
@@ -1,5 +1,9 @@
 # Changelog for `proto-lens-protoc`
 
+## Pending
+- Fix a potential naming conflict when message types and enum values
+  are the same except for case.
+
 ## v0.5.0.0
 
 ### Breaking Changes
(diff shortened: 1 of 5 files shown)


blackgnezdo:
Looks like an old ghc bug?

ghc: panic! (the 'impossible' happened)
  (GHC version 8.2.1 for x86_64-unknown-linux):
	getUnboxedSumName
  513
  Call stack:
      CallStack (from HasCallStack):
        prettyCurrentCallStack, called at compiler/utils/Outputable.hs:1133:58 in ghc:Outputable
        callStackDoc, called at compiler/utils/Outputable.hs:1137:37 in ghc:Outputable
        pprPanic, called at compiler/prelude/KnownUniques.hs:104:5 in ghc:KnownUniques
Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug

PR #319, opened 2019-05-02
In `protoc` install commands, also extract `include/`

jarollz:
Also unzip `include/` folder into `usr/local/include` to make sure plugins that commonly requires them works immediately.


judah:
CLA is good: I pushed a minor change to this PR.

PR #320, opened 2019-05-03
Fix a bug in text format parsing

kgraney:
It's valid for text format protobufs to contain strings with unescaped quote
characters different from the characters used to delimit the string.


judah:
Thank you for  the fix!

PR #324, opened 2019-06-09
fix #61: implement decodeDelimitedMessageH (decoding from a file handle)

### LINKED ISSUE #61: How to parse delimited messages

ulysses4ever:


judah:
Thank you for the patch!  I'll take a look once the CLA is signed.  Let me know if you have any questions about getting it working.

mboes:
@ulysses4ever as been added to Tweag's google-contributors@ whitelist. So he has transitively signed the CLA.

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
```diff
@@ -76,14 +80,43 @@ putBytes = Builder.byteString
...
-            then return $! n'
-            else loop (128*s) n'
+        getVarIntLoopFinish n s b loopStart
+
+-- Same as getVarInt but reads from a Handle
+getVarIntH :: Handle -> IO Word64
+getVarIntH h = do
+    buf <- malloc
+    let loopStart !s !n = do
+          _ <- hGetBuf h buf 1
```
judah: You should check the result of `hGetBuf`.  It might return a value of zero at EOF, in which case `b` would be undefined.  At worst case, that would mean this could loop forever.

proto-lens-tests/package.yaml
```diff
@@ -209,6 +209,17 @@ tests:
       - Proto.Any
       - Proto.Any_Fields
 
+  decode_delimited:
```
judah: Nit: please follow the convention of the other tests and name this "decode_delimited", and name the corresponding source file "decode_delimited_test.hs".

proto-lens-tests/tests/decode_delimited.hs
```diff
@@ -0,0 +1,41 @@
...
+main :: IO ()
+main = testMain
+    [ testCase "buildDelimited/decodeDelimited" $ do
+          let foo = defMessage & a .~ 42 & b .~ "hello" :: Foo
+          let bs = runBuilder . buildMessageDelimited $ foo
+          B.writeFile filename bs
+          h <- openBinaryFile filename ReadMode
+          any1 <- decodeDelimitedMessageH h :: IO (Either String Foo)
+          hClose h
+          removeIfExists filename
```
judah: When would we not expect the file to exist?  Would it make the test simpler/more robust if we just threw an error if the file unintentionally disappeared?

proto-lens-tests/tests/decode_delimited.hs
```diff
@@ -0,0 +1,41 @@
...
+  where handleExists e
+          | isDoesNotExistError e = return ()
+          | otherwise = throwIO e
+
+main :: IO ()
+main = testMain
+    [ testCase "buildDelimited/decodeDelimited" $ do
+          let foo = defMessage & a .~ 42 & b .~ "hello" :: Foo
+          let bs = runBuilder . buildMessageDelimited $ foo
+          B.writeFile filename bs
```
judah: Let's use the `temporary` package to avoid writing files in the current directory, which would be the `proto-lens-tests` source directory in this case.
http://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withSystemTempFile

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -7,8 +7,11 @@ module Data.ProtoLens.Encoding (
     -- ** Delimited messages
     buildMessageDelimited,
     parseMessageDelimited,
+    decodeDelimitedMessageH,
```
judah: I think we should try to mirror the naming scheme of the other functions:

`decodeMessage`, `decodeMessageDelimited`, `decodeMessageDelimitedH`.

Does that sound reasonable to you?

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
ulysses4ever: Good point! What would be a reasonable way to fail gracefully in case `hGetBuf` returns `0`?

proto-lens-tests/tests/decode_delimited.hs
ulysses4ever: That makes sense. Do you want me to leave just the call to `removeFile` (which might throw) without additional bookkeeping.

proto-lens/src/Data/ProtoLens/Encoding.hs
ulysses4ever: Ouch, silly me! I can't believe I got this wrong. Will fix.

proto-lens-tests/package.yaml
ulysses4ever: Sorry, I didn't get: is the name of the test wrong too? The name of the _file_ I will fix, thanks!

judah: Yeah, by convention the test names in package.yaml also have "_test" at the end.

proto-lens-tests/tests/decode_delimited.hs
judah: `withSystemTempFile` should remove the file automatically, so I believe you could just delete the call.

proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
judah: Hm, after thinking about it: since the function that ultimately calls this one returns `IO (Either String ...)`, you should probably return a similar type from this function: `IO (Either String Word64)`.

Maybe it would make sense to simplify the internal error-passing via `ExceptT String IO`.  Up to you.

ulysses4ever:
I tried to address all your comments, thanks a lot for reviewing! Waiting for more comments!

proto-lens/src/Data/ProtoLens/Encoding.hs
```diff
@@ -47,3 +52,10 @@ parseMessageDelimited = do
     len <- Bytes.getVarInt
     bytes <- Bytes.getBytes $ fromIntegral len
     either fail return $ decodeMessage bytes
+
+-- | Same as @decodeMessage@ but for delimited messages read through a Handle
+decodeMessageDelimitedH :: Message msg => Handle -> ExceptT String IO msg
```
judah: Hm, my original suggestion had been to keep ExceptT *internal* to the implementation, and still return `IO (Either String msg)` from this function.

Now that I see this, I'm on the fence which way is better, but leaning towards plain `IO`.  I think `ExceptT` could make sense if we had several different functions we wanted to compose, but I'm not sure we'll add anything beyond this one.

What do you think?  Would `ExceptT` simplify the code you plan to use this with, or would you end up just unwrapping it with `runExceptT` when you used it?

ulysses4ever: I agree with you: returning IO of Either feels more natural in the current interface than returning ExceptT. Let me implement that.

ulysses4ever: Done.

ulysses4ever:
Thanks a lot! I'd appreciate if you could release it any time soon (but no rush, of course).

judah:
@ulysses4ever I've released proto-lens-0.5.1.0 containing this change.

ulysses4ever:
@judah thanks a lot!

PR #331, opened 2019-08-30
Add `methodOptions` to `HasMethodImpl` to provide custom method options.

judah:
We provide the whole proto message, rather than pulling off individual fields.  The set of fields may change for different builds of proto-lens (or based on extensions, though we can't see those directly due to #27).

Internally, we save the proto message in the source file as an encoded string, and decode it into a Haskell type lazily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,8 +111,6 @@ will generate the haskell files `Proto/Project/{Foo,Bar}.hs`.
 - Extensions (proto2-only) are not supported.
 - Unknown proto2 enum values cause a decoding error, instead of being preserved
   round-trip.
-- Files with `import public` statements compile correctly, but don't explicitly
-  reexport the definitions from those imports.
 
 # Troubleshooting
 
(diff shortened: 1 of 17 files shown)


judah:
Fixed; PTAL.

It turns out the test (which relied on GHC's overlapping pattern matching) wasn't catching errors correctly:
https://github.com/google/proto-lens/blob/6e8b73d6a29556cc2e3550a9235d7b7df42b7a20/proto-lens-tests/tests/service_test.hs#L16

It was worse on newer GHC, but even on older ones it missed a misuse of `"bidiStreaming"` instead of `"biDiStreaming":
https://github.com/google/proto-lens/blob/6e8b73d6a29556cc2e3550a9235d7b7df42b7a20/proto-lens-tests/tests/service_test.hs#L74

I fixed it by just calling `evaluate` on the values to force GHC to resolve the constraints.

PR #336, opened 2019-09-07
Use ghc-source-gen instead of haskell-src-exts

judah:
The translation is mostly a straight translation between the types
in haskell-src-exts and in GHC/ghc-source-exts.  For example, `Name`
becomes `OccNameStr` and `QName` becomes `RdrNameStr`.

One significant difference between the APIs is that `ghc-source-exts`
doesn't define IsString instances for expressions and types, instead
requiring use of the `var` or `bvar` functions.  Again, the translation
is straightforward.

Also adds an extra "CommentedModule" type to describe the whole
module with top-level (Haddock) comments.  Currently, `ghc-source-gen` doesn't
support them well, because GHC's pretty-printer doesn't do a good job
rendering them.  (Of note, `ghc-exactprint` also keeps track of comments
manually.)


judah:
Note: this PR is diffbased on #334 and #335.

judah:
I just did a quick test (current version on the `ghc-8.8` branch).  It seems it won't be too much work (in part because ghc-source-gen already supports the newer compiler). I ran into a couple issues, both of which were minor:

- `Data.ProtoLens.Encoding.Parser` needs to update for `MonadFail`
- `Data.ProtoLens.Setup` needs to be updated for `Cabal-3`.

judah:
Actually, it looks like test-framework doesn't build with ghc-8.8 due to regex-base not building.  It looks like the package is somewhat stale, but may be taken over soon:
https://mail.haskell.org/pipermail/libraries/2019-September/029927.html

PR #337, opened 2019-09-08
Drop support for ghc-8.0/Cabal<2 and clean up CI build versions.

judah:
- Removes support for ghc-8.0 (base-4.9) and Cabal < 2.0
- Bumps main stack.yaml to lts-14 (ghc-8.6.5)
- Adds CI coverage for newer versions of Cabal (non-stack)


judah:
ghc-8.2.1 was released more than two years ago, so it's reasonable for us to drop support for older GHC.

PR #340, opened 2019-09-23
Update installing-protoc.md

steve-prod:
Fix URL.


blackgnezdo:
Thank for the poke, the URLs are slightly different, so updated and submitted.

PR #353, opened 2019-10-20
Remove a hpack hack around generated modules.

judah:
It's fine to have the .cabal files require Cabal>=2.0 because
proto-lens-setup has that restriction now, as of #337.


blackgnezdo:
I like the red, but CI disagrees.

PR #357, opened 2019-10-23
Reduce the API of proto-lens-protoc.

### LINKED ISSUE #347: Remove the library from proto-lens-protoc
proto-lens-protoc exposes an internal library, which has not been used in practice.  It limits our ability to make changes inside of it without requiring a major version bump.

Currently the library serves another purpose, namely, to make proto-lens-setup "depend: on the proto-lens-protoc executable (via the dep on the library).  However, that doesn't work with cabal new-build.

We should just remove the library (and proto-lens-setup's dependency on it), and require using build-tool-depends instead as in #212.

judah:
Previously we exposed most of the internals as a library, with the hope
that they'd be useful for plugins (i.e. for services).  But now we're
exposing service information at the type level, and the library
continues to be unused.  Removing most of it allows us to change
the internals without requiring a major version bump.  For example,
with this change the migration to ghc-source-gen would have only
required a minor version bump.

Fixes #347.  I think it's not feasible to completely remove the library
at this time, since we still want to share logic around module
names between proto-lens-protoc and proto-lens-setup.


blackgnezdo:
Did you search hackage for reverse deps to declare the library unused in practice?

judah:
Yeah, they're all only using it for the codegen, not for libraries.
https://packdeps.haskellers.com/reverse/proto-lens-protoc

PR #358, opened 2019-10-26
Move descriptor/plugin bindings to proto-lens-protobuf-types.

### LINKED ISSUE #344: Collision with code generated from Google Protobuf repo
Hi,

I'm attempting to use `proto-lens` with [rules_haskell](https://github.com/tweag/rules_haskell) to generate a Haskell library for [Open Match](https://github.com/googleforgames/open-match) protobuf definitions, and I'm running into a situation where the `.proto` file depends on `google/protobuf/descriptor.proto`, which means that in my `BUILD.bazel` file I must depend on a `@com_google_protobuf//:descriptor_proto` target so that `protoc` can find the files that are included.

This is a pattern in other repos using Bazel, and you can see an example here: https://github.com/googleapis/googleapis/blob/master/google/api/BUILD.bazel#L12

Example of a file I'm attempting to generate Haskell from: https://github.com/googleforgames/open-match/blob/master/third_party/google/api/annotations.proto#L20

Trying to do so will cause `proto-lens-protoc` to generate a `Proto.Google.Protobuf.Descriptor` module which collides with the one included with `proto-lens` since https://github.com/google/proto-lens/pull/140 .

```
   Ambiguous module name `Proto.Google.Protobuf.Descriptor':
     it was found in multiple packages:
       descriptor-proto proto-lens-0.4.0.1
   |
25 | import qualified Proto.Google.Protobuf.Descriptor
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I haven't been able to find a solution that would make both `protoc` and `proto-lens` happy. Any suggestion would be appreciated.

Thanks!

judah:
Previously the checked-in bootstrap protos were part of the proto-lens package.
After this change, they're included in the `other-modules` of
proto-lens-protoc`.  Instead, the exposed versions of the modules are
provided by proto-lens-protobuf-types, just like the other proto
files.

Fixed #344, which describes why this was causing issues with some
build systems like Bazel.

Another side effect of this change is: previously the bootstrap protos
had special logic for how they imported modules.  After this change
they can depend on them from proto-lens-runtime, just like other
generated protos.  This simplifies some logic in proto-lens-protoc.

Also update the bootstrap script.  It's mostly the same as before,
except that we first checkout an old version of the bootstrap proto
bindings that will be compatible with the version of proto-lens we're
bootstrapping against. (They'll be overwritten at the end of the
script anyway.)

### DIFF
--- a/proto-lens-protoc/app/Proto/Google/Protobuf/Compiler/Plugin.hs
+++ b/proto-lens-protoc/app/Proto/Google/Protobuf/Compiler/Plugin.hs
@@ -7,31 +7,31 @@ module Proto.Google.Protobuf.Compiler.Plugin (
         CodeGeneratorRequest(), CodeGeneratorResponse(),
         CodeGeneratorResponse'File(), Version()
     ) where
-import qualified Control.DeepSeq
-import qualified Data.ProtoLens.Prism
-import qualified Prelude
-import qualified Data.Int
-import qualified Data.Monoid
-import qualified Data.Word
-import qualified Data.ProtoLens
-import qualified Data.ProtoLens.Encoding.Bytes
-import qualified Data.ProtoLens.Encoding.Growing
-import qualified Data.ProtoLens.Encoding.Parser.Unsafe
-import qualified Data.ProtoLens.Encoding.Wire
-import qualified Data.ProtoLens.Field
-import qualified Data.ProtoLens.Message.Enum
-import qualified Data.ProtoLens.Service.Types
-import qualified Lens.Family2
-import qualified Lens.Family2.Unchecked
-import qualified Data.Text
-import qualified Data.Map
-import qualified Data.ByteString
-import qualified Data.ByteString.Char8
-import qualified Data.Text.Encoding
-import qualified Data.Vector
-import qualified Data.Vector.Generic
-import qualified Data.Vector.Unboxed
-import qualified Text.Read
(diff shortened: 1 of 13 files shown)


proto-lens-protoc/app/Proto/Google/Protobuf/Compiler/Plugin.hs
```diff
@@ -7,31 +7,31 @@ module Proto.Google.Protobuf.Compiler.Plugin (
...
-import qualified Data.Text
-import qualified Data.Map
-import qualified Data.ByteString
-import qualified Data.ByteString.Char8
-import qualified Data.Text.Encoding
-import qualified Data.Vector
-import qualified Data.Vector.Generic
-import qualified Data.Vector.Unboxed
-import qualified Text.Read
+import qualified Data.ProtoLens.Runtime.Control.DeepSeq as Control.DeepSeq
```
blackgnezdo: Is this right?

judah: Yeah, this is intentional.  Now that the module lives in proto-lens-protoc, it can depend on proto-lens-runtime (which reexposes the `Data.ProtoLens.Runtime.*` modules), the same as regular generated proto modules.

PR #359, opened 2019-10-26
Bump the version to 0.6.

judah:
Packages like proto-lens-optparse and proto-lens-arbitrary that
are compatible with the older versions only get a minor bump.

### DIFF
--- a/proto-lens-arbitrary/Changelog.md
+++ b/proto-lens-arbitrary/Changelog.md
@@ -1,7 +1,8 @@
 # Changelog for `proto-lens-arbitrary`
 
-## Pending
+## v0.1.2.8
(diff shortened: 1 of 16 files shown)


judah:
Note: #358 requires a major version bump from proto-lens.

proto-lens-arbitrary/Changelog.md
```diff
@@ -1,7 +1,8 @@
 # Changelog for `proto-lens-arbitrary`
 
-## Pending
+## v0.1.2.8
 - Bump lower bounds to base-4.10 (ghc-8.2).
+- Bump upper bounds for proto-lesn-0.6.
```
blackgnezdo: lens

judah: Done.

PR #361, opened 2019-11-09
Add messageDescriptor to Message.

dustin:
Each compiled proto has a DescriptorProto associated with it that was
previously lost during the compilation phase.  This DescriptorProto is
sometimes needed for interop when you want a message to be able to
describe itself as a type before providing serialized values.


proto-lens-tests/tests/imports_test.hs
```diff
@@ -59,3 +62,16 @@ testUseBootstrapped :: TestTree
...
     testField @Imports.UseBootstrapped @Plugin.CodeGeneratorRequest #request
+
+testDescriptor :: TestTree
+testDescriptor = testCase "testDescriptor" $ do
+  d <- case descProto (Proxy :: Proxy ImportsTransitive.TransitiveDep) of
+         Left x -> assertFailure x
+         Right x -> pure x
+
+  "TransitiveDep" @=? view #name d
+  2 @=? (length . view #field) d
```
dustin: I'm not sufficiently familiar with lens to write the thing I  actually wanted.

     ["x", "y"] @=? views #field . something . #name

Something dumb like this works:

    map (\x -> x ^. #name) $ r ^. #field

proto-lens-tests/tests/imports_test.hs
```diff
@@ -59,3 +62,16 @@ testUseBootstrapped :: TestTree
 testUseBootstrapped = testCase "testUseBootstrapped" $ do
     testField @Imports.UseBootstrapped @Descriptor.DescriptorProto #descriptor
     testField @Imports.UseBootstrapped @Plugin.CodeGeneratorRequest #request
+
+testDescriptor :: TestTree
```
dustin: I  wasn't sure whether having this test here makes sense, or whether I should make a new test just for this functionality.

proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -925,6 +931,8 @@ messageInstance env protoName m =
     , valBind "buildMessage" $ generatedBuilder m
     ]
   where
+    msgDescriptor = fmap (int . toEnum . fromEnum) . BS.unpack . runBuilder . buildMessage $
```
judah: You can just use `encodeMessage` instead of `runBuilder . buildMessage`.

proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -904,6 +908,8 @@ messageInstance :: Env RdrNameStr -> T.Text -> MessageInfo OccNameStr -> [RawIns
 messageInstance env protoName m =
     [ funBind "messageName" $ match [wildP] $
           var "Data.Text.pack" @@ string (T.unpack protoName)
+    , funBind "messageDescriptor" $ match [wildP] $
+          var "Data.ByteString.pack" @@ list msgDescriptor
```
judah: I think it would be better to use OverloadedStrings (which this file already supports), e.g.

```
messageDescriptor _ = "abcde"
```

The benefits would be: (1) the syntax is more compact, and (2) under the hood, GHC optimizes away the intermediate string/list and stores the packed bytestring directly in the generated code.

I think it would be something like:

```
funBind "messageDescriptor" $ match [wildP] $ string msgDescriptor
...
  where
    msgDescriptor = fmap (toEnum . fromEnum) . BS.unpack . buildMessage $
       messageDescriptor m
```

(which avoids needing to add an import of Data.ByteString.Char in the generated code).

proto-lens-tests/tests/imports_test.hs
judah: Let's make a new test. (Maybe something like `message_test` or `descriptor_test`?)  It seems like no existing test is a good fit for this, and it's generally better not to use the same sample "data" for two different uses.  (Also, in this case, transitive_dep.proto comes from a different test package, making it a little more complicated than needed.)

proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -904,6 +908,8 @@ messageInstance :: Env RdrNameStr -> T.Text -> MessageInfo OccNameStr -> [RawIns
 messageInstance env protoName m =
     [ funBind "messageName" $ match [wildP] $
           var "Data.Text.pack" @@ string (T.unpack protoName)
+    , funBind "messageDescriptor" $ match [wildP] $
```
judah: On second thought, maybe we should call this `packedMessageDescriptor` since it returns a ByteString?  That would avoid a name clash if we want to define a function returning the descriptor itself (see my comment about proto-lens-protobuf-types).

proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate.hs
dustin: OverloadedStrings wasn't active in the generated protos.  I was trying to follow along with  the `messageName` thing which is using `Data.Text.unpack`.  I updated this to use a string, though.

dustin:
I didn't add `Data.ProtoLens.Descriptor` as I wasn't quite sure how to resolve the dependency.  I think I did the rest of the things, though.

Issue #364, opened 2019-11-10
QuickCheck failure in proto-lens-discrimination


judah:
We got a failure in CI for an unrelated change:

```
QuickCheck found a failure:

Log files have been written to: /home/travis/build/google/proto-lens/.stack-work/logs/

Test suite failure for package proto-lens-discrimination-0.1.1.1

    discrimination_test:  exited with: ExitFailure 1

Full log available at /home/travis/build/google/proto-lens/.stack-work/logs/proto-lens-discrimination-0.1.1.1-test.log

    field value:
      int32: [OK]
      string: [OK]
      enum Bar pos: [OK]
      enum Bar neg: [OK]
      enum Bar eq: [OK]
      message Foo eq: [OK]
      message Foo lt: [OK]

    map:
      empty: [OK]
      empty LT singleton: [OK, passed 100 tests]
      k => v1 `compare` k => v2 = v1 `compare` v2: [OK, passed 100 tests]
      k1 => v `compare` k2 => v = k1 `compare` k2: [OK, passed 100 tests]
      same as Ord: [OK, passed 100 tests]

    compares by field: [OK, passed 100 tests]

    Message:
      defMessage EQ defMessage: [OK]
      lower tags compared first: [OK]
      symmetry: [OK, passed 100 tests]
      transitivity: [Failed]

    *** Failed! Falsified (after 49 tests):

    ArbitraryMessage {unArbitraryMessage = {baz: BAZ2 bar_defaulted: BAR5 baz_defaulted: BAZ2 true_defaulted: true false_defaulted: false}}
    ArbitraryMessage {unArbitraryMessage = {baz: BAZ2 bar_defaulted: BAR5 baz_defaulted: BAZ2 true_defaulted: true false_defaulted: false}}
    ArbitraryMessage {unArbitraryMessage = {baz: BAZ2 bar_defaulted: BAR5 baz_defaulted: BAZ2 true_defaulted: true false_defaulted

: false}}

    (used seed 8805600212509756313)

             Properties  Test Cases   Total       
     Passed  6           10           16          
     Failed  1           0            1           
     Total   7           10           17
```

judah:
CC @awpr

awpr:
Oh, it looks like the test is wrong: when all three messages are equal, the greatest and least under sorting are still equal.  Apparently it's rare enough for QuickCheck to generate three equal messages that it hasn't been encountered until now.

judah:
[...]

PR #365, opened 2019-11-10
Fix transitivity property test of message Sorts.  (#364)

awpr:
If all three messages are equal, they'll get sorted into some arbitrary order, and the least won't be less than the greatest.  So, assert that either the least is less than the greatest, or all three are equal.

Note: I wasn't able to convince proto-lens to build in a few tens of minutes (all the .proto files in submodules were missing), so I gave up and decided to let the CI deal with it.  So, this might not even type-check yet.

### DIFF
--- a/proto-lens-discrimination/package.yaml
+++ b/proto-lens-discrimination/package.yaml
@@ -46,6 +46,7 @@ tests:
     source-dirs: tests
     dependencies:
       - HUnit >= 1.3 && < 1.7
+      - QuickCheck >= 2.8 && < 2.14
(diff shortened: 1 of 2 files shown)


judah:
Thanks for the fix!  For future reference, you can use `git submodule update --init --recursive` as described here:

https://github.com/google/proto-lens#building-from-head

awpr:
Ah, this used to be much more difficult in the past, so I gave up too easily out of fear of how bad it used to be.  Actually pretty easy now!  Fixed a missing dependency on QuickCheck.

PR #367, opened 2019-11-12
Adding list of reserved keywords for field names.

wigahluk:
Also adding some minor format fix

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -1,7 +1,6 @@
 # Scalar Type Reference
 
-Reference for mappings between scalar protobuf types and generated Haskell types.
-
+This is the reference for mappings between scalar protobuf types and generated Haskell types.
 
 | Protobuf type  | Haskell type               |
 | -------------- | -------------              |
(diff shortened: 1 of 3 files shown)


docs/types.md
```diff
@@ -1,8 +1,7 @@
-
-This page serves as a reference for mappings between scalar protobuf types and generated Haskell types.
-
 # Scalar Type Reference
 
+Reference for mappings between scalar protobuf types and generated Haskell types.
```
judah: I would prefer a full sentence here rather than just the sentence fragment.  (But the change you made to move it after the title looks correct.

wigahluk: Done

wigahluk:
I can help on that doc re-org if you want, do you have a specific example in mind? I'm asking because it seems that these files are merged into a github page at "google.github.io/proto-lens/" which is pretty close to a single doc.

judah:
The one thing I think we're missing is a short guide to exactly what lenses get generated for which kinds of proto constructs.  The tutorial contains some of that, but it currently focuses more on the internals of the generated record (which are supposed to be opaque anyway) rather than the lenses.

As you say, it probably doesn't make a difference whether there's a single doc or a set of docs linked from our github.io page.

PR #369, opened 2019-11-28
Add FileDescriptorProto to generated protos.

dustin:
This change adds packedFileDescriptor to each Message as well as a
fileDescriptor helper function for getting the FileDescriptorProto
from this Message.

The FileDescriptorProto is required for reflectively understanding all
of the contents of a proto.  It's used, for example, by APIs that
persist protos.

While it may seem a little odd to have each Message defined in a proto
return the same FileDescriptorProto, this is about the third thing I
tried working with and found the alternatives to be much worse.

1. If `packedFileDescriptor` is exported from the module itself, much
   code is broken due to duplicate definitions on flat imports.
2. Adding a new module, e.g. `MyProto_Descriptor` worked around the
   issue at no cost by having the flat space separated.

However, both of these mean that an API that needs access to the
FileDescriptorProto has no way to discover it from a proto it's
given.  Manual registration of FileDescriptorProto would always be
required ahead of time for any API that needs it, and this is rather
cumbersome and difficult to enforce programmatically.

### DIFF
--- a/proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate.hs
+++ b/proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate.hs
@@ -39,6 +39,7 @@ import Text.Printf (printf)
 
 import Proto.Google.Protobuf.Descriptor
     ( EnumValueDescriptorProto
+    , FileDescriptorProto
     , FieldDescriptorProto
     , FieldDescriptorProto'Type(..)
     )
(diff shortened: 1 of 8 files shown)


dustin:
I finally got around to looking at this last night.  I've not tested that this does all of the things I need to do in my actual application as it's a little hard to transplant all of this in place, but I think the functionality is important regardless, so  I figured I'd submit for review.

I actually did three different versions of this (as described in the commit message) and this one made the most sense to me, but I'm open to other suggestions.

proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate.hs
```diff
@@ -140,6 +141,17 @@ generateModule modName imports publicImports definitions importedEnv services
         , info <- allMessageFields env m
         , inst <- recordFieldLenses info
         ]
+    -- The packedFileDescriptorProto is a file level definition that's
+    -- shared across all message definitions.  If there are no
+    -- messages, it's omitted.
```
judah: Maybe clarify comment: it's omitted because it's not used?  (It took me a second to figure this out)

dustin:
Ah, I had style-haskell on save and got everything working and then had to rewrite it all to avoid making a large mess.  I guess my test run lied to me.  I see the failures now.  A couple lines didn't make it into the last rewritten thing.  I've fixed this up now (and updated that comment).

PR #370, opened 2019-12-05
Fix a warning when there are only enums, no messages.

judah:
Previously we got:

```
...../build/enum_test/autogen/Proto/EnumOnly.hs:72:1: error: [-Wunused-top-binds, -Werror=unused-top-binds]
    Defined but not used: ‘packedFileDescriptor’
   |
72 | packedFileDescriptor
   | ^^^^^^^^^^^^^^^^^^^^
```

The check for whether the file had any messages was incorrect, since it didn't account for enums which are tracked in the same map.

### DIFF
--- a/proto-lens-tests/tests/enum_test.hs
+++ b/proto-lens-tests/tests/enum_test.hs
@@ -17,6 +17,7 @@ import Proto.Enum
     , Alias(..)
     , ManyCasesProto
     )
+import Proto.EnumOnly (Lone(..))
 import Proto.Enum_Fields
 import Data.Function (on)
 import Data.ProtoLens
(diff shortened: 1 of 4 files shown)


proto-lens-tests/tests/enum_test.hs
```diff
@@ -163,3 +165,5 @@ testAliases = testCase "alias" $ do
 
 testManyCases =
     runTypedTest (roundTripTest "many cases" :: TypedTest ManyCasesProto)
+
+testLone = testCase "no messages" $ LONE @?= LONE
```
jinwoo: I'm not sure I understand this test. Isn't `LONE` always `== LONE`?

judah: Added a comment.  The test is intentionally trivial; the goal is just to make this file import the other module to make sure it compiles properly.

jinwoo: Got it. Thanks for the comment.

PR #376, opened 2020-03-24
Add support for message set wire format

ntnj:
This can be enabled by adding "option message_set_wire_format = true" to
the proto message.

The encoding format is taken from C++ code
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wire_format.cc#L1046

### DIFF
--- a/proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate/Encoding.hs
+++ b/proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate/Encoding.hs
@@ -368,7 +368,9 @@ unknownFieldCase ::
         loop (over unknownFields (\!t -> y:t) x) ...
 -}
 unknownFieldCase info loop x = match [wire] $ do' $
-    [ strictP y <-- var "Data.ProtoLens.Encoding.Wire.parseTaggedValueFromWire" @@ wire
+    [ strictP y <-- if messageDescriptor info ^. #options ^. #messageSetWireFormat
+        then var "Data.ProtoLens.Encoding.Wire.parseMessageSetTaggedValueFromWire" @@ wire
+        else var "Data.ProtoLens.Encoding.Wire.parseTaggedValueFromWire" @@ wire
     ]
     ++
     [ stmt $ case' y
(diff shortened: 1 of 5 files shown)


ntnj:
I wanted to check if this solution looks good before I add any unit tests. Thanks.

blackgnezdo:
@judah will you have time to look at this or should I?

judah:
Thank you for your contribution!  I had a little bit of difficulty understanding the purpose from the code you linked, so I wanted to double-check.  Am I right that this (a) only affects extensions, and (b) since proto-lens doesn't currently support accessing extensions (#27), it only currently matters for unknown fields?  Also, I'm curious if you're able to give more details of your use case, given that this encoding has been deprecated for a long time?

Overall, this change looks reasonable to me given how simple it is. As you mentioned, please add a new unit test in `proto-lens-tests` (probably just a new one in `UnknownFields.hs`/`unknown_fields.proto`).  Please also add a comment in Data.ProtoLens.Encoding.Wire describing the format and its use; maybe just add a link to here: https://github.com/protocolbuffers/protobuf/blob/dec4939439d9ca2adf2bb14edccf876c2587faf2/src/google/protobuf/descriptor.proto#L444

ntnj:
Yes, this only affects extensions. Since they're not supported, I've been able to use them by manually encoding them into ```unknownFields``` by using ```TaggedValue tag (Lengthy $ encodeMessage msg``` (see the unit test for example)

This encoding has been deprecated, but is still used at Google in a commonly used message for interoperability with proto1.

My parse method (parseMessageSetTaggedValueFromWire) expects the field number first and then the actual message and then end tag, but C++ implementation supports reading them in either form and is generally more failure resilient in case of unexpected inputs. Should I handle these cases or is this minimal handling fine? I've added a test which shows the unsupported parse which would have worked with C++ implementation. I checked the implementations of C++ and Go encoder, and any encoded messages generated by those implementations should be parse-able without handling these extra cases.

judah:
That makes sense.  Especially with the tests, we can change it later if we need to support the more general version.

judah:
@blackgnezdo this passed CI, and I'm not sure why the status hasn't updated:
https://travis-ci.org/github/google/proto-lens/builds/668209311
I think I don't have enough permissions to merge it manually though.

blackgnezdo:
Surprisingly I also see a greyed-out button. Do you know how to kick-off another CI run?

judah:
I just pushed an empty commit.  It looks like the CLA bot is now confused, but I think you can just trigger it manually or add the "cla: yes" label (I believe it won't let me because I'm not in the google org).

judah:
@googlebot I consent.

PR #378, opened 2020-04-09
Update Coffee-order tutorial README.md

championballer:
Corrected the link to the order.proto in the readme of the coffee-order tutorial


blackgnezdo:
Thanks for the fix!

PR #382, opened 2020-05-27
Optimize putVarInt

ethercrow:
Some evidence of the redundancy of `.&. 127`:
```
> quickCheck (\(x :: Word64) -> (fromIntegral (x .|. 128) :: Word8) == (fromIntegral (x .&. 127 .|. 128) :: Word8))
+++ OK, passed 100 tests.
```

### DIFF
--- a/proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
+++ b/proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
@@ -126,7 +126,7 @@ testMsb b = (b .&. 128) /= 0
 putVarInt :: Word64 -> Builder
 putVarInt n
     | n < 128 = Builder.word8 (fromIntegral n)
-    | otherwise = Builder.word8 (fromIntegral $ n .&. 127 .|. 128)
+    | otherwise = Builder.word8 (fromIntegral $ n .|. 128)
                       <> putVarInt (n `shiftR` 7)
 
 getFixed32 :: Parser Word32
--- a/proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
+++ b/proto-lens/src/Data/ProtoLens/Encoding/Bytes.hs
@@ -123,11 +123,17 @@ decodeVarIntStep n s b = n + s * fromIntegral (b .&. 127)
 testMsb :: Word8 -> Bool
 testMsb b = (b .&. 128) /= 0
 
+{-# INLINE putVarInt #-}
 putVarInt :: Word64 -> Builder
-putVarInt n
-    | n < 128 = Builder.word8 (fromIntegral n)
-    | otherwise = Builder.word8 (fromIntegral $ n .|. 128)
-                      <> putVarInt (n `shiftR` 7)
+putVarInt n | n < 128 = Builder.word8 (fromIntegral n)
+putVarInt n =
+  -- Note that the narrowing done by (fromIntegral :: Word64 -> Word8) 
+  -- effectively throws out bits higher than 8 so we don't have to do
+  -- an explicit (n .&. 127)
+  Builder.word8 (fromIntegral n .|. 128) <> go (n `shiftR` 7)
+  where
+  go x | x < 128 = Builder.word8 (fromIntegral x)
+  go x = Builder.word8 (fromIntegral x .|. 128) <> go (x `shiftR` 7)
 
 getFixed32 :: Parser Word32
 getFixed32 = getWord32le


ethercrow:
@googlebot I signed it!

blackgnezdo:
Did you by chance look at core? I'm curious if there's any change in generated code.

ethercrow:
@blackgnezdo I failed to build the project at all. Both cabal and stack complain about `unexpected ","` in proto-lens-runtime.cabal

ethercrow:
Another `putVarInt` idea is:

```
putVarInt n | n < 128 = word8 (fromIntegral n)
putVarInt n = putVarIntRecursive n
{-# INLINE putVarInt #-}

putVarIntRecursive = ... current implementation ...
```

So that GHC can optimize code like `putVarInt <some-small-constant>`, I have lots of it in https://github.com/ethercrow/lightstep-haskell/blob/master/gen/Proto/Collector.hs

But again it's pure speculation that it doesn't. I'm only assuming the lack of inlining because the function is recursive.

ethercrow:
I looked at godbolt.org with `-O2`, it shows the difference of

```
        orq $128,%rax
```

vs

```
        andl $127,%eax
        orq $128,%rax
```

judah:
Thanks @ethercrow for the patch! This change looks reasonable to me. One small request: can you please add a comment indicating that fromIntegral does the "and" operation for us? I think that would make it a little easier to compare to other implementations of the encoding.

ethercrow:
@judah Sure, I added a comment.

I also went ahead with the inlining change. Couldn't get tests to run locally, but hopefully Travis will tell me if everything's fine.

judah:
@ethercrow, for your second proposed change, I think we'll want to run more detailed analysis (actual benchmarking and looking at Core) to confirm that it's a net improvement. I'm also wondering whether `INLINABLE` would be more appropriate than `INLINE`. Maybe we should split it out into a second PR?

To benchmark the change you ought to be able to run `stack bench proto-lens-benchmarks --ghc-options=-O2`. (You'll also need to run the Setup instructions here: https://github.com/google/proto-lens#setup) Generally, running the benchmark before/after a change and comparing the results can give some useful information (although it's sometimes a little noisy due to GC, local machine fluctuations, etc.)

If that doesn't work for you, and you're interested in continuing to work on it, please let us know the exact error message you're getting and we can help diagnose it further.

ethercrow:
I got `stack test` and `stack bench` working, here are the encoding numbers:

This PR (but with INLINABLE) | This PR (with INLINE) | Master

Screen Shot 2020-05-28 at 9 59 18 PM

Apply grains of salt as needed.

judah:
Thanks for those benchmark results! It does look like we get some significant improvements on int32-packed,  (~5.8us-> ~2.0us), no-unused (57us -> 21us) and with-unused (60us -> 21us), which is expected since their encodings are particularly tag-heavy. And it seems INLINE is necessary rather than INLINABLE.

I did see int32-packed get *slower* in speed, which is a bit unexpected since we'd expect the bulk of the time to be copying the packed bytes. Possibly that was just noise though.

Once concern I have is that the INLINE causes more code to be generated and compilation time go up (for example, when varint-encoding an arbitrary int). As an experiment I hacked proto-lens-protoc to only build Proto.Google.Protobuf.Descriptor, and saw that compiling it was consistently ~16% slower: 38s at HEAD vs 44s with this change. I'm guessing this is due to the extra inlining making the generated code noticeably larger.

@blackgnezdo , any thoughts?

blackgnezdo:
Does compile time change uniformly across GHC versions? Is there a known GHC issue which could address slow compilation?

I don't want to make a trade-off between compilation speed and runtime speed for the users. I also don't like extra knobs and this one will probably be on the uglier side (CPP?).

One possible data point would be to try on the change on a bigger sample of protos. I'll try to find a willing internal volunteer to see what the impact is for us.

ethercrow:
Just my unsolicited opinion: if compilation speed is important to your organization, please contribute some developer time or sponsor some existing devs who optimize GHC in their free time. Seeing issues like https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3363 I wonder how many more of such small things are waiting there just because so few people are looking for them.

ethercrow:
Is there hope for this PR?

ethercrow:
Weekly bump.

judah:
@ethercrow I apologize for the delay in reply.  Unfortunately I personally won't have much bandwidth to investigate the compilation time slowdown in the near term. Also, as I'm not at Google anymore I don't have access to their internal protos to do more thorough testing.

I think your first change (removing `.&. 127`) is uncontroversial; how about we merge that one for now? I opened #396 for that.

judah:
Incidentally, here's one idea that I had, but didn't have time to try out. Our use of `putVarInt` with constants is all in *generated* code (output from `proto-lens-protoc`). See for example:
https://github.com/google/proto-lens/blob/0ac7a36e91038ce40a9e4108f36e75960281224e/proto-lens-protoc/app/Data/ProtoLens/Compiler/Generate/Encoding.hs#L556

So instead of relying on the INLINE mechanism, we could make your proposed transformation entirely within the code generation stage. More precisely, everywhere that we currently use `putVarInt' @@ int ...` (which generates the Haskell code `putVarInt {n}`) we could replace with a function that special-cases it to generate `putWord8 {n}`. For example, somethng like:

```
putVarIntConst :: Integer -> HsExpr'
putVarIntConst n
  | n < 128 = var "putWord8" @@ int n
  | otherwise = var "putVarInt" @@ int n
```

That should have essentially no impact on compilation times, and compile down to the same code as your proposed patch.

ethercrow:
> So instead of relying on the INLINE mechanism, we could make your proposed transformation entirely within the code generation stage.

👍

PR #394, opened 2020-06-04
Support enums in type registry.

kmandelbaum:
This can be used to parse things like package.name.Enum.Type in a
context where the enum type is not known a priori.
This change to proto-lens requires re-generating modules with the updated version of proto-lens-protoc.


kmandelbaum:
@googlebot I fixed it.

PR #400, opened 2020-11-19
Support proto3 optional

### LINKED ISSUE #380: Field presence support per proto3 change
Should we do something about newly allowed `optional` fields in proto3?
https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md

gagern:
Fixes #380 following [instructions](https://github.com/protocolbuffers/protobuf/blob/e2cc2de/docs/implementing_proto3_presence.md) recommended in https://github.com/google/proto-lens/issues/380#issuecomment-619569960.

The main change here is to take the proto3 descriptor message but convert it to an internal representation more in line with what we'd generate for proto2. We treat the field as an `OptionalMaybeField` and drop the synthetic `oneof`. The serialization already works out of the box, since a `oneof` member is converted into an `OptionalMaybeField` as well.

For this to work, the descriptor proto needs to be updated to include the `FieldDescriptorProto.proto3_optional` field. We also need the `FEATURE_PROTO3_OPTIONAL` enum constant available so this plugin can tell `protoc` that it does support the new feature.

### DIFF
--- a/proto-lens-protoc/app/protoc-gen-haskell.hs
+++ b/proto-lens-protoc/app/protoc-gen-haskell.hs
@@ -18,6 +18,7 @@ import Data.Semigroup ((<>))
 import qualified Data.Set as Set
 import qualified Data.Text as T
 import Data.Text (Text, pack)
+import Data.Bits
 import Data.ProtoLens (defMessage, decodeMessage, encodeMessage)
 -- Force the use of the Reflected API when decoding DescriptorProto
 -- so that we can run the test suite against the Generated API.
@@ -27,6 +28,7 @@ import Lens.Family2
 import Proto.Google.Protobuf.Compiler.Plugin
     ( CodeGeneratorRequest
     , CodeGeneratorResponse
+    , CodeGeneratorResponse'Feature(..)
     )
 import Proto.Google.Protobuf.Descriptor (FileDescriptorProto)
 import System.Environment (getProgName)
@@ -63,7 +65,10 @@ makeResponse dflags prog request = let
     header f = "{- This file was auto-generated from "
                 <> (f ^. #name)
                 <> " by the " <> pack prog <> " program. -}\n"
+    features = [CodeGeneratorResponse'FEATURE_PROTO3_OPTIONAL]
     in defMessage
+           & #supportedFeatures .~
+               (foldl (.|.) zeroBits $ fmap (toEnum . fromEnum) features)
            & #file .~ [ defMessage
                             & #name .~ outputName
                             & #content .~ outputContent
--- a/proto-lens-tests/tests/proto3_optional_test.hs
+++ b/proto-lens-tests/tests/proto3_optional_test.hs
@@ -3,15 +3,28 @@
 -- Use of this source code is governed by a BSD-style
 -- license that can be found in the LICENSE file or at
 -- https://developers.google.com/open-source/licenses/bsd
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 module Main (main) where
 
 import Proto.TestProto3Optional
 import Proto.TestProto3Optional_Fields
+import Data.Int
 import Data.ProtoLens
-import Lens.Family2 ((&), (.~), (^.))
+import Data.ProtoLens.Encoding.Parser (Parser, runParser)
+import Data.ProtoLens.Field (HasField, fieldOf)
+import Data.ProtoLens.Labels ()
+import Data.Proxy
+import Lens.Family2 ((&), (.~), (^.), (^..))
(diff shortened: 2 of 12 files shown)


blackgnezdo:
Very cool! @judah if you have cycle you'd be the best reviewer. If not I'll do my best.

proto-lens-protoc/app/protoc-gen-haskell.hs
```diff
@@ -63,7 +65,10 @@ makeResponse dflags prog request = let
     header f = "{- This file was auto-generated from "
                 <> (f ^. #name)
                 <> " by the " <> pack prog <> " program. -}\n"
+    features = [CodeGeneratorResponse'FEATURE_PROTO3_OPTIONAL]
     in defMessage
+           & #supportedFeatures .~
+               (foldl (.|.) zeroBits $ fmap (toEnum . fromEnum) features)
```
judah, line 71: Could we simplify this to `supportedFeatures .~ toEnum (fromEnum FEATURE_PROTO3_OPTIONAL)` for now? We can add the `foldl` logic later once we have more than one feature to report.

proto-lens-tests/tests/test_proto3_optional.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package proto3_optional;
+
+message Foo {
+  int32 untracked = 1;
+  optional int32 tracked = 2;
+  oneof nonsynth {
```
judah, line 8: I didn't see nonsynth or alternative used in any of the tests; could we remove them? (There are already other tests specifically focused on oneofs.)

proto-lens-tests/tests/proto3_optional_test.hs
```diff
@@ -0,0 +1,49 @@
...
+    , testGroup "lens"
+        [ testCase "unset" $ defFoo ^. tracked @=? 0
+        , testCase "default" $ (defFoo & tracked .~ 0) ^. tracked @=? 0
+        ]
+    , testGroup "maybe'lens"
+        [ testCase "unset" $ defFoo ^. maybe'tracked @=? Nothing
+        , testCase "default" $ (defFoo & tracked .~ 0) ^. maybe'tracked
+          @=? Just 0
+        ]
+    -- TODO(gagern): Check that the generated file doesn't contain
```
judah, line 45: Maybe hack this by just defining `type {NAME} = ()` where `NAME` is the oneof type that we would have otherwise generated?

Up to you how much effort (if any) to put into this.

judah (changes requested):
Thank you for implementing this! I really appreciate your helpful comments explaining the new logic.

I have a handful of minor requests, but otherwise your approach looks good to me.

proto-lens-tests/tests/test_proto3_optional.proto
gagern: The compile-time checks I added to guard against extraneous generated API make use of these additional symbols.

proto-lens-protoc/app/protoc-gen-haskell.hs
gagern: I would argue not to. Performance isn't an issue here. From a simplicity point of view, the `foldl` expression is certainly longer, but it does convey the message that this is an or-ed bit field without looking at the proto comments. So in my opinion it makes reading easier, since I as a reader would expect this to be multi-valued and surprised to not see a list. Also surprised to see the conversion from enum to int, which makes little sense for a single value but much for a bit field. The readers most likely to actually care about this code would be those adding another feature, in which case not having to come up with the `foldl` will save time.

Have I convinced you? If not, I will simplify.

proto-lens-tests/tests/proto3_optional_test.hs
gagern: Good idea. I've written something like this, but with the `HasField` class a single added symbol isn't good enough, so it ended up being closer to the full generated code than your example suggests.

gagern:
I invested some more time trying to get tests to verify compile time failures. I managed to call GHC as library, but failed to give it all the library locations it would need since I don't know how to get them from stack. I [asked on Stack Overflow](https://stackoverflow.com/q/64933993/1468366) for suggestions on how to achieve this. I don't suggest blocking this pull request waiting for that since the tests as they are now should be sufficient in the short run. It's only in the long run when APIs change that the process of manually enabling `#if` blocks becomes unreliable.

judah (approved):

judah:
This looks good to me. Thank you again for implementing this feature!

gagern:
@judah According to internal view, the google-cla bot is talking to you, waiting for your consent. I wonder whether there is any specific reason it doesn't say so in the public wording. Thanks for catching the outdated version numbers.

judah:
@googlebot I consent.

judah:
@blackgnezdo the status report from Travis CI isn't being seen by GitHub, even though the tests passed:
https://travis-ci.org/github/google/proto-lens/pull_requests

It looks like we'll need to update the protections on the branch as described in the "Remedy" section here:
https://github.com/travis-ci/travis-ci/issues/10204

I don't have permission to change that, probably because I'm not in the Google org anymore.

judah:
@blackgnezdo I took a closer look and still wasn't able to figure it out. Maybe ask the Google GitHub admins?

For now, since I've verified the build status manually, I'll temporarily disable the branch protection so I can merge this.

PR #402, opened 2020-12-05
Update upper bounds for QuickCheck and bytestring

lehins:


lehins:
I am not going to use my gmail with github

PR #408, opened 2021-02-28
fix source directory in person tutorial

stuebinm:
in the package.yml snipped given in the Readme of the person tutorial,
the build directory was set to ".", which will fail with a Main.hs created
in the `src` directory (as described in the tutorial).

The actual code in the same directory appears to be fine.

(just noticed this while trying to follow the tutorial)


stuebinm:
Huh. I don't particularly feel like creating a google account and signing the agreement just for three new bytes in a readme, so i guess just take this PR as a bug report …

PR #409, opened 2021-05-06
Add ServiceDescriptor to generated services.

ntnj:
This is similar to generated messageDescriptor in #361.

### DIFF
--- a/proto-lens/src/Data/ProtoLens/Service/Types.hs
+++ b/proto-lens/src/Data/ProtoLens/Service/Types.hs
@@ -23,8 +23,10 @@ module Data.ProtoLens.Service.Types
   , StreamingType (..)
   ) where
 
+import qualified Data.ByteString as B
 import Data.Kind (Constraint)
 import Data.ProtoLens.Message (Message)
+import Data.Proxy (Proxy(..))
 import GHC.TypeLits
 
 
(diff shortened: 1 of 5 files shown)


blackgnezdo:
We still don't have a CI, so I tested that this works locally.

proto-lens/src/Data/ProtoLens/Service/Types.hs
```diff
@@ -53,6 +55,7 @@ class ( KnownSymbol (ServiceName s)
   type ServicePackage s :: Symbol
   type ServiceMethods s :: [Symbol]
 
+  packedServiceDescriptor :: Proxy s -> B.ByteString
```
judah, line 58: I think it would be useful to add an analogue to the `messageDescriptor` function from `proto-lens-protobuf-types`:
http://hackage.haskell.org/package/proto-lens-protobuf-types-0.7.0.0/src/src/Data/ProtoLens/Descriptor.hs

```
serviceDescriptor :: forall s . Service s => DescriptorProto
```

judah (approved):
LGTM with one suggestion.

Incidentally I think this will require a major version bump due to the change in the class definition, but that can be done at the next release of this package.

proto-lens/src/Data/ProtoLens/Service/Types.hs
ntnj: Thanks for the review. I've added a commit to implement this.

blackgnezdo:
Great thanks for the change!

PR #411, opened 2021-05-16
Support all valid text format boolean field values

### LINKED ISSUE #410: Text format parser doesn't accept all valid boolean values
Ran into this at work. Unfortunately the text format specification doesn't seem to be public. 😞 But you can see from the C++ parser code that the valid values for a true boolean field are 1, "true", "True", or "t" (and similar for a false value): https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/text_format.cc#L775. Whereas the proto-lens parser currently only accepts 1 or "true": https://github.com/google/proto-lens/blob/master/proto-lens/src/Data/ProtoLens/TextFormat.hs#L320.

Of course this is very easy to fix. I'll probably get around to doing it myself later today, assuming nothing gets in the way.

agrue:
I'm not sure the text format is exactly "standardized", but this is the set supported by the official C++ parser (see #410).

Fixes #410

### DIFF
--- a/proto-lens/src/Data/ProtoLens/TextFormat.hs
+++ b/proto-lens/src/Data/ProtoLens/TextFormat.hs
@@ -317,8 +317,8 @@ makeScalarValue BoolField (Parser.IntValue x)
 makeScalarValue DoubleField (Parser.DoubleValue x) = Right x
 makeScalarValue FloatField (Parser.DoubleValue x) = Right (realToFrac x)
 makeScalarValue BoolField (Parser.EnumValue x)
-    | x == "true" = Right True
-    | x == "false" = Right False
+    | x `elem` ["true", "True", "t"] = Right True
+    | x `elem` ["false", "False", "f"] = Right False
     | otherwise = Left $ "Unrecognized bool value " ++ show x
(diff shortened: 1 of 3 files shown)


proto-lens/src/Data/ProtoLens/TextFormat.hs
```diff
@@ -317,8 +317,8 @@ makeScalarValue BoolField (Parser.IntValue x)
 makeScalarValue DoubleField (Parser.DoubleValue x) = Right x
 makeScalarValue FloatField (Parser.DoubleValue x) = Right (realToFrac x)
 makeScalarValue BoolField (Parser.EnumValue x)
-    | x == "true" = Right True
-    | x == "false" = Right False
+    | x `elem` ["true", "True", "t"] = Right True
```
blackgnezdo, line 320: The "1" and "0" aren't worth the test?

blackgnezdo (approved):

proto-lens/src/Data/ProtoLens/TextFormat.hs
agrue: "1" and "0" are handled differently; looks like they're parsed to `Int` first (see the `Parser.IntValue` match a few lines up).

PR #428, opened 2021-10-10
Fixed stack-9.0.yaml

jchia:
- nightly resolver needs full name with date
- cassava is not a dependency of any of the packages, though an indirect
  dependency. In any case, the removed cassava flag is not applicable because
  nightly-2021-10-07 has bytestring-0.10.12.1.
- ghc-source-gen is in nightly-2021-10-07 so extra-deps is unnecessary
- allow-newer is unnecessary


blackgnezdo (approved):
Perfect, thanks @jchia

PR #430, opened 2021-11-22
Fix the source-dirs to reflect Main.hs location

blackgnezdo:


proto-lens-tutorial/person/README.md
```diff
@@ -123,7 +123,7 @@ Then, we'll put it in an `executable` section to `package.yaml` that specifies i
 executables:
   person:
     main: Main.hs
-    source-dirs: '.'
+    source-dirs: 'src'
```
jinwoo, line 126: Should this actually be a list of dirs, as written in the package.yaml file?

```
source-dirs:
  - src
```

jinwoo: Just a single entry seems to work fine. LGTM.

jinwoo (approved):

PR #431, opened 2021-11-30
Add GHC 9.2 support

brandon-leapyear:


.travis.yml
```diff
@@ -31,6 +31,8 @@ matrix:
     addons: {apt: {packages: [libgmp-dev]}}
   - env: BUILD=stack STACK='stack --stack-yaml=stack-9.0.yaml'
     addons: {apt: {packages: [libgmp-dev]}}
+  - env: BUILD=stack STACK='stack --stack-yaml=stack-9.2.yaml'
```
blackgnezdo, line 34: We switched to github actions for CI. Please update  `.github/workflows/stack-ci.yml`.

Could you do me favor and delete `.travis.yml`? It only confuses us...

blackgnezdo (changes requested):

.travis.yml
brandon-leapyear: Seems like I didn't need to change anything in `stack-ci.yml`? Just `matrix.yaml`? (Not sure why matrix is defined in another file; why not just define as a normal matrix in the github actions config?)

blackgnezdo (approved):

blackgnezdo:
Thanks @brandon-leapyear!

brandon-leapyear:
@blackgnezdo when do you think you'll get around to making a release with 9.2 support?

blackgnezdo:
You are right, it's about time. I'm retesting with ghc-9.2.2 now and should be able to publish this week.

brandon-leapyear:
Thanks!

PR #433, opened 2022-04-05
Relax text bounds to allow text-2.0

blackgnezdo:
This was tested by using an explicit override with this diff:
```
--- a/stack-9.2.yaml
+++ b/stack-9.2.yaml
@@ -21,5 +21,10 @@ packages:
 - proto-lens-tests
 - proto-lens-tests-dep

+extra-deps:
+  - text-2.0
+  - Cabal-3.6.3.0@sha256:ff97c442b0c679c1c9876acd15f73ac4f602b973c45bde42b43ec28265ee48f4,12459
+  - parsec-3.1.15.0@sha256:a162d4cc8884014ba35192545cad293af0529fe11497aad8834bbaaa3dfffc26,4429
+
 ghc-options:
   "$locals": -Wall -Werror
```

### DIFF
--- a/proto-lens-arbitrary/Changelog.md
+++ b/proto-lens-arbitrary/Changelog.md
@@ -1,5 +1,8 @@
 # Changelog for `proto-lens-arbitrary`
 
+## v0.1.2.10
+- Relax upper bounds for ghc-9.2
+
 ## v0.1.2.9
 - Bump upper bound to allow base-4.14.
 
--- a/proto-lens-arbitrary/package.yaml
+++ b/proto-lens-arbitrary/package.yaml
@@ -18,7 +18,7 @@ dependencies:
   - base >= 4.10 && < 4.17
   - bytestring >= 0.10 && < 0.12
   - containers >= 0.5 && < 0.7
-  - text == 1.2.*
+  - text >= 1.2 && < 2.1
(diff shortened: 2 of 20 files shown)


blackgnezdo:
text-2.0 will be in ghc-9.4, so I might as well front-load this boundary relaxation since I'm in the area.

blackgnezdo:
@brandon-leapyear if you have some time, could you check I'll cover everything by releasing these packages:
 -  proto-lens
 - proto-lens-arbitrary
 - proto-lens-optparse
 - proto-lens-protobuf-types
 - proto-lens-protoc
 - proto-lens-runtime
 - proto-lens-setup

brandon-leapyear:
At the very least, that's all that my company uses :). A quick search of Hackage indicates that that seems correct

brandon-leapyear:
FYI @blackgnezdo `proto-lens` is unique in that most other packages don't need to be released for us to be able to use it; `stack.yaml` allows specifying github repos with commits as dependencies instead of going to Hackage. But Stack has issues doing the same with the `proto-lens` git repo: https://github.com/commercialhaskell/stack/issues/5603

So in general, releasing `proto-lens` is more urgent than other packages because it's impossible to depend on it unless it's on Hackage

jinwoo (approved):
LGTM

PR #434, opened 2022-04-05
Re-number -arbitrary -optparse -runtime -setup

blackgnezdo:
They previously had no Changelog entries and it was confusing.
All of these are live on Hackage now.


blackgnezdo:
This should be the real ghc92 release tag.

PR #440, opened 2022-08-21
Allow GHC 9.4 (and CI) for `proto-lens` and `proto-lens-runtime`

ysangkok:
With a little subsumption related change, `proto-lens` can be made to work with GHC 9.4.

Other packages have dependency trees that don't support GHC 9.4 yet, so I deliberately made this PR small to just add support to these two packages mentioned in the title.

Tested using

    cabal test -w ghc-9.4.1 --constraint='vector>=0.13'

### DIFF
--- a/.github/workflows/cabal-ci.yml
+++ b/.github/workflows/cabal-ci.yml
@@ -1,0 +1,19 @@
+on:
+  - push
+  - pull_request
+name: cabal build with ghc 9.4
+jobs:
+  runhaskell:
+    name: cabal test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - uses: haskell/actions/setup@v2
+        with:
+          ghc-version: '9.4.2'
+          cabal-version: '3.8.1.0'
+      - run: cabal build --dependencies-only --project-file ghc94.cabal.project all
+      - run: cabal build                     --project-file ghc94.cabal.project all
+      - run: cabal test                      --project-file ghc94.cabal.project all
--- a/ghc94.cabal.project
+++ b/ghc94.cabal.project
@@ -1,0 +1,3 @@
+ packages:
+   proto-lens
+   proto-lens-runtime
--- a/proto-lens-runtime/package.yaml
+++ b/proto-lens-runtime/package.yaml
@@ -16,7 +16,7 @@ extra-source-files:
 
 library:
   dependencies:
-    - base >= 4.10 && < 4.17
+    - base >= 4.10 && < 4.18
     - bytestring >= 0.10 && < 0.12
     - containers >= 0.5 && < 0.7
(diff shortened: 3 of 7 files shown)


blackgnezdo:
Thanks for the fixes @ysangkok! I wonder how hard it would be to add CI support for 9.4.

ysangkok:
@blackgnezdo Is there a way to run the tests on CI using Cabal instead of Stack? Because there isn't a Stackage snapshot for GHC 9.4 yet, and it could take a while. I think having a Cabal build could be useful in the future too, not only for 9.4.

blackgnezdo:
> @blackgnezdo Is there a way to run the tests on CI using Cabal instead of Stack? Because there isn't a Stackage snapshot for GHC 9.4 yet, and it could take a while. I think having a Cabal build could be useful in the future too, not only for 9.4.

If you know how to set up a cabal-based build in addition to what we have, that would be great.

ysangkok:
@blackgnezdo I added a Cabal build that uses [haskell-CI/haskell-ci](https://github.com/haskell-CI/haskell-ci). It will run in parallel to the existing workflow.

.github/workflows/cabal-ci.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          submodules: true
+      - uses: haskell/actions/setup@v2
+        with:
+          ghc-version: '9.4.1'
+          cabal-version: '3.8.1.0'
+      - run: cat cabal.project | grep -E '^packages|lens$|runtime$' > new.cabal.project
+      - run: mv new.cabal.project cabal.project
+      - run: cabal build all
+      - run: cabal test proto-lens
```
parsonsmatt, line 20: An alternative here is to have a `ghc94.cabal.project` and use `cabal build --project-file ghc94.cabal.project all`. That way this step can be checked in and maintained separately.

ysangkok: Oh, great! I have applied your suggestion.

blackgnezdo:
@jinwoo when you get a second, could you merge this if the builds are all green?

blackgnezdo:
Hmm, I wonder if ghc 9.4.2 is too new somehow?

ysangkok:
Yes, it is indeed too new. Let's wait for

* https://github.com/haskell/ghcup-metadata/pull/31

ysangkok:
@blackgnezdo GHC 9.4.2 was added to the GHCup metadata and the test [now passes](https://github.com/ysangkok/proto-lens/runs/8028172879) if rerun. Doesn't seem like I have the privileges to rerun it.

jinwoo:
I re-triggered it. I'll merge if it succeeds.

jinwoo (approved):

jinwoo:
Merged. Thanks for the contribution!

parsonsmatt:
Could this be released on Hackage soon?

PR #455, opened 2023-12-02
Extend proto-lens API to support serializing the protobuf JSON format

iand675:
I needed to support JSON deserialization to the Protobuf models to integrate with a third party system with a large number of protocol buffer definitions, so it was infeasible to hand-write them.

This PR adds support for the serialization formats outlined here: https://protobuf.dev/programming-guides/proto3/#json

I also bumped some dependencies and replaced the `Setup.hs` codegen approach for the official protobuf types because the Setup.hs approach absolutely refused to work in the upstream Nix project. I could probably roll it back, but it's nice IMO to be able to use some of the base packages to the ecosystem without having to have the protocol buffer compiler installed.

### DIFF
--- a/proto-lens-json/src/Data/ProtoLens/Json.hs
+++ b/proto-lens-json/src/Data/ProtoLens/Json.hs
@@ -18,7 +18,12 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 -- | ToJSON/FromJSON instances for protocol buffer Messages to use with aeson.
-module Data.ProtoLens.Json where
+module Data.ProtoLens.Json 
+  ( JsonMessage(..)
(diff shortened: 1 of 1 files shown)


blackgnezdo:
Thanks for the change @iand675.

You are right that the bootstrap story is a complicated one. We pondered it quite a bit before landing on this one. The reasoning was that people who want to work with protobuf wouldn't consider it a chore to have `protoc` available. Keeping the generated files comes with a maintenance cost. I'd rather we limit the scope of your change to the json related features you added.

I'm also semi-blocked on `ghc-source-gen` ghc 9.6 support and the lack of CI support for the newer GHC versions in this project. I'd be happy if these were resolved.

WutherHeights (changes requested):
Does not pass the required checks.

PR #456, opened 2023-12-15
Allow GHC 9.4 in more packages

### LINKED ISSUE #447: Support GHC 9.4
The latest ghc is 9.4.4 but the upper bound is <9.3.
The latest base is 4.17.0.0 but the upper bound is <4.17.

Consequently, many packages in this project are not in stackage nightly.

ysangkok:
Now that ghc-source-gen supports GHC 9.4, we can support it here too.

Tested using

    cabal test -c 'text>=2.1' -w ghc-9.4

in proto-lens-protoc. With cabal-install v3.10.2.0.

Fixes #447.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -11,8 +11,3 @@ packages:
   proto-lens-setup
   proto-lens-tests
   proto-lens-tests-dep
-
-source-repository-package
-  type: git
-  location: https://github.com/google/ghc-source-gen.git
-  tag: master
--- a/cabal.project
+++ b/cabal.project
@@ -11,3 +11,8 @@ packages:
   proto-lens-setup
   proto-lens-tests
   proto-lens-tests-dep
+
+source-repository-package
+  type: git
+  location: https://github.com/google/ghc-source-gen.git
+  tag: master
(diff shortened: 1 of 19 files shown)


cabal.project
```diff
@@ -11,3 +11,8 @@ packages:
   proto-lens-setup
   proto-lens-tests
   proto-lens-tests-dep
+
+source-repository-package
```
blackgnezdo, line 15: I understand why you want to make this change, but I'd prefer to have the new version of ghc-source-gen pushed to hackage instead. 

The rest of the changes are good to go, if you want them merged first.

blackgnezdo (changes requested):
Thank you for the patch!

cabal.project
ysangkok: I have removed these lines as requested.

blackgnezdo:
@ysangkok could you check if https://github.com/google/proto-lens/pull/457 subsumes this PR? I will need to do the change similar to your most recent commit, but otherwise we should be able to jump all the way to 9.6 now.

ysangkok:
@blackgnezdo yeah, i've check for subsumption now and left a couple of comments

ysangkok:
closing in favour of https://github.com/google/proto-lens/pull/457

PR #460, opened 2024-01-09
Add support for additional well known types

domMayhew:
Some well known types with message definitions in Protobuf source code were not being converted into Haskell. Proto definitions that depend on those types (e.g., FieldMask) failed to build. This PR adds those types to the auto-generated code so proto definitions that depend on them can build.

There is an open PR (#455) that overlaps with this one. However, that PR has been asked to remove the changes that overlap with this one. This PR simply adds to the existing `Setup.hs` approach.


blackgnezdo (approved):
Thank you for the PR, looks good to me!

Issue #468, opened 2024-04-22
Support "Protobuf Editions"


agrue:
Protobuf Editions overview: https://protobuf.dev/editions/overview/

This feature introduces a new syntax (e.g. `edition = "2023";`) that currently breaks `proto-lens`. It crashes in this partial function here: https://github.com/google/proto-lens/blob/a64671071c1790c3e94eba3fd289ded2e44a74a4/proto-lens-protoc/app/Data/ProtoLens/Compiler/Definitions.hs#L104

The error is `Unknown syntax type "editions"`.

chungyc:
I have looked at this sufficiently closely by now that I know what needs to be done.

1. Update the bootstrapping generated code for the protocol buffer message types used by `protoc` plugins to a recent version, so that it includes the fields necessary to support Protobuf Editions.  The update will not affect current behavior.  This will be handled by this pull request: #469

2. Update the code generator to tell `protoc` that it can support Protobuf Editions, which a subsequent change can do: https://github.com/chungyc/proto-lens/pull/1

   Depending on the preference of the reviewers, I can modify this change to say it only supports `EDITION_LEGACY`, which would allow us to merge in code partially supporting Protobuf Editions early without allowing it to process Protobuf Editions files before the support is complete.  Or I can make all the changes (i.e, including the subsequent steps) to support Protobuf Editions in a single much larger pull request.

3. Implement [feature resolution](https://protobuf.dev/editions/implementation/#feature-resolution) for Protobuf Editions.

4. Plumb through Protobuf Editions [feature sets] throughout the code generator, replacing adjustments based on syntax type with those based on feature sets.  We would basically replace [`SyntaxType`] with feature sets, with the addition of overriding features at a local scope as needed.

[feature sets]: https://github.com/protocolbuffers/protobuf/blob/bd3629c6b9b501df1d1c9e4813c4158062a3aeb7/src/google/protobuf/descriptor.proto#L967
[`SyntaxType`]: https://github.com/google/proto-lens/blob/a64671071c1790c3e94eba3fd289ded2e44a74a4/proto-lens-protoc/app/Data/ProtoLens/Compiler/Definitions.hs#L96

These changes to the `protoc` plugin should be all that is necessary to support Protobuf Editions 2023, as it is limited to being a stepping stone from proto2 and proto3 to Protobuf Editions instead of adding any new functionality.

PR #469, opened 2024-09-07
Catch up on changes in protobuf up to v28.0.

chungyc:
* Update location to the current one at https://github.com/protocolbuffers/protobuf
* Update `protobuf` submodule to v28.0
* Regenerate bootstrapping protobuf bindings with `runghc bootstrap.hs`

This is preparatory work for supporting Protobuf Editions (issue #468), which requires the updates to the protocol buffer message types defined for `protoc` plugins.


blackgnezdo (approved):
Looks good, thanks @chungyc

PR #471, opened 2024-09-13
Check out protoc v28.x during Stack continuous integration.

chungyc:
We will need to use a much more recent version of `protoc` if we wish to test Protobuf Editions files.  Old `protoc` versions will not be able to parse new syntax.  For example, a newer version is required for the tests in https://github.com/chungyc/proto-lens/pull/8 to pass.

### DIFF
--- a/.github/workflows/stack-ci.yml
+++ b/.github/workflows/stack-ci.yml
@@ -78,7 +78,7 @@ jobs:
       - name: Install Protoc
         uses: arduino/setup-protoc@v1
         with:
-          version: '3.x'
+          version: '28.x'
 
       # Install Stack without GHC first, so we have an opportunity to cache the
       # Pantry package index.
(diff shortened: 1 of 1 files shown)


.github/workflows/stack-ci.yml
```diff
@@ -76,9 +76,9 @@ jobs:
 
       # Installs the recent protoc, for now unpinned to gain some experience
       - name: Install Protoc
-        uses: arduino/setup-protoc@v1
```
chungyc, line 79: Mayhaps v1 doesn't support downloading v28.0.  Updated to the latest v3.

chungyc: This indeed seems to have been the case.  `protoc` installs successfully when testing in a pull request within my own fork: https://github.com/chungyc/proto-lens/pull/10

blackgnezdo (approved):

blackgnezdo:
Thanks for the changes @chungyc !

PR #473, opened 2024-09-16
Support feature resolution for Protobuf Editions.

chungyc:
This implements Protobuf Editions [feature resolution](https://protobuf.dev/editions/implementation/#feature-resolution) for the Haskell `protoc` plugin.

This embeds the binary protobuf message that `protoc` outputs for feature set defaults, which is deserialized at runtime.  It is done this way since it does not seem sustainable to manually translate the protobuf message to Haskell code every time there is a new edition.  It also seems to be the [same approach](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/cpp_edition_defaults.h) that protocolbuffers/protobuf intends for use with custom code generators written in C++.  This is not stored in a separate file with the binary protobuf message to avoid adding additional dependencies such as [file-embed](https://hackage.haskell.org/package/file-embed).

Tests are not included in this pull request, but this code is tested in a subsequent change (https://github.com/chungyc/proto-lens/pull/8), which confirm that the feature resolution works by maintaining the same behavior for proto2 and proto3.  This is due to the fact that tests cannot directly import these modules.  Perhaps the modules in `app/` should be turned into an [internal library](https://cabal.readthedocs.io/en/3.4/cabal-package.html#sublibs) for unit tests, but that change is not for this one.

This does not complete Protobuf Editions support, which requires two more changes: https://github.com/chungyc/proto-lens/pull/7 (control proto2 and proto3 behavior based on features, not syntax type), https://github.com/chungyc/proto-lens/pull/8 (override features in enclosed scopes)

For https://github.com/google/proto-lens/issues/468.


chungyc:
Are they any concerns that need to be addressed?

I can also have a pull request with all the changes for Protobuf Editions support at once if that is the preference.

blackgnezdo:
I was hoping @agrue would take the lead on this one :) I'm a bit distanced from protobufs these days (I know!) so it's harder for me to find the time to invest into a proper review.

agrue:
Sure, I will take a look. Just need to find the time.

proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Defaults.hs
```diff
@@ -0,0 +1,45 @@
...
+{-| 'FeatureSetDefaults' message containing feature defaults.
+
+This contains the defaults from editions @LEGACY@ to @EDITION_2023@
+for the native features defined by @google.protobuf.FeatureSet@.
+-}
+nativeDefaults :: FeatureSetDefaults
+nativeDefaults
+  | Right m <- msg = m
+  | Left e <- msg = error $ "unable to decode built-in defaults: " ++ e
+  where msg = decodeMessage serializedNativeDefaults
```
agrue, line 30: Why not just build the proto message with regular Haskell code rather than decode a binary string embedded in a source file...?

proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Features.hs
```diff
@@ -0,0 +1,86 @@
...
+otherwise the field in B is used as is.
+
+Consider using this function as an infix operator.
+For example,
+
+>>> let c = a `mergedInto` b
+
+could be read like "message C is the message A merged into message B".
+
+If merging was generally supported for all messages,
```
agrue, line 73: Sounds like a nice feature request for proto-lens (maybe it's been requested already, I haven't checked).

proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Defaults.hs
agrue: I see, you addressed this in the PR description:

> it does not seem sustainable to manually translate the protobuf message to Haskell code every time there is a new edition

Hm, really? The proto is pretty small, at least right now. My concern is that it makes really hard to tell what the values actually are (e.g. for debugging). I think it's better to have the proto generated by Haskell code, but it's up to you.

agrue (approved):

proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Defaults.hs
chungyc: "Not sustainable" was really just another way of saying "an annoying amount of manual work".  I would rather trust the output of `protoc` instead of including complicated instructions to generate the equivalent Haskell code.

(Truth be told, I wasn't actually able to get the correct incantations for either `ghci` or `stack ghci` to even attempt a `show` for the `FeatureSetDefaults` message, assuming that `show` for protobuf messages even outputs valid Haskell code.)

proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Features.hs
chungyc: There is no such feature request yet.  Maybe someday we can add it by making protobuf messages monoids ...

PR #474, opened 2024-09-21
Set up Dependabot.

chungyc:
The GitHub actions used by proto-lens are becoming years out of date.  We should let automation handle preparing updates, even if approval remains manual.

Also use grouping to reduce the number of pull requests.


blackgnezdo (approved):
Thanks!

ysangkok:
I've added Haskell support to Renovate by the way. Renovate also checks Cabal bounds, unlike Dependabot:

https://discourse.haskell.org/badges/18/nice-topic?username=janus

PR #476, opened 2024-09-30
Add HLint workflow.

### LINKED ISSUE #279: Add hlint to proto-lens's CI
As suggested here: https://github.com/google/proto-lens/pull/275#discussion_r241504249
Run hlint in CI against 

We may want to avoid doing this for every single LTS release, and instead pin to a single hlint version.  (Maybe: do it in a separate job.)

chungyc:
Adds a code scanning workflow using HLint.  Suggestions and warnings are included inline for pull requests and the code scanning dashboard in the security tab.  See https://github.com/haskell-actions/hlint-scan#examples for an example.

This sets up HLint to ignore warnings that would be triggered by the existing code.  We would likely want to enable some of them, for example, `Redundant $` to remove unnecessary `$` operators.  Others could be marked as permanently disabled.  This could be done in future changes.

Resolves #279.

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -1,0 +1,12 @@
+name: HLint
+on: [push, pull_request]
+
+jobs:
+  hlint:
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload results to GitHub code scanning.
+      security-events: write
+    steps:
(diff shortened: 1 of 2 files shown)


blackgnezdo (approved):

blackgnezdo:
Looks awesome, thanks for finding this gem!

PR #477, opened 2024-10-01
Use per-edition feature defaults instead of syntax type to control behavior.

chungyc:
This basically replaces `SyntaxType` distinguishing between proto2 and proto3 with `FeatureSet` controlling individual features in Protobuf Editions.  For example, instead of checking whether the syntax is proto3 when determining whether repeated fields should use packed encoding by default, we check whether `features.repeated_field_encoding` is `PACKED`.  The default feature set [compatible with proto2 and proto3](https://protobuf.dev/editions/features/#preserving) are provided by `protoc` and included in `Data.ProtoLens.Compiler.Defaults`.  The existing tests continue to pass, showing that the features have been determined correctly.

This does not complete support for Protobuf Editions.  This requires the ability to override options in certain scopes, e.g., file scope or field scope, which will be added in https://github.com/chungyc/proto-lens/pull/8.

This is the second to last change required for https://github.com/google/proto-lens/issues/468.

### DIFF
--- a/proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Features.hs
+++ b/proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Features.hs
@@ -48,7 +48,7 @@ for a particular edition.
 featuresForEditionFromDefaults :: FeatureSetDefaults -> Edition -> FeatureSet
 featuresForEditionFromDefaults defaults edition
   | (d : _) <- candidates = (d ^. #overridableFeatures) `mergedInto` (d ^. #fixedFeatures)
-  | otherwise = defMessage
+  | otherwise = error $ "Unsupported edition with tag number: " ++ show (fromEnum edition)
   where
     candidates = dropWhile (\d -> d ^. #edition > edition) recentFirst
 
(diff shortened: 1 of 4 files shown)


chungyc:
If you want to take a look, @agrue, here is the next change.

proto-lens-protoc/app/Data/ProtoLens/Compiler/Editions/Features.hs
```diff
@@ -48,7 +48,7 @@ for a particular edition.
 featuresForEditionFromDefaults :: FeatureSetDefaults -> Edition -> FeatureSet
 featuresForEditionFromDefaults defaults edition
   | (d : _) <- candidates = (d ^. #overridableFeatures) `mergedInto` (d ^. #fixedFeatures)
-  | otherwise = defMessage
+  | otherwise = error $ "Unsupported edition with tag number: " ++ show (fromEnum edition)
```
agrue, line 51: I would rather not use `error`. Can you work some better error handling in here?

chungyc: Changed this to return an `Either Text a`, and the call chain has been updated accordingly all the way up to `makeResponse`, which will now include the error string in the response.

agrue (approved):

agrue:
Thank you.

PR #479, opened 2024-10-02
Override features defined in enclosed scopes.

chungyc:
This completes support of Protobuf Editions to the same level that proto-lens supports proto2 and proto3 except for the following, which I do not think are critical bugs:

* `features.utf8_validation = NONE` does not disable UTF-8 validation.

* If more than one field sharing the same message type are in a single message type, and if one of the fields has `features.message_encoding = DELIMITED`, then all of these fields will use `DELIMITED` encoding.

Unevaluated tests are intentionally included despite warnings by HLint so that they are still built and can be easily enabled later if the above are fixed.

Resolves https://github.com/google/proto-lens/issues/468.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,15 @@ will generate the haskell files `Proto/Project/{Foo,Bar}.hs`.
 - Unknown proto2 enum values cause a decoding error, instead of being preserved
   round-trip.
 
+## Protobuf Editions
+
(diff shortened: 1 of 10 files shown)


chungyc:
@agrue, this last change completes support for Protobuf Editions.

agrue (approved):
Thanks for all the tests!

PR #489, opened 2024-10-10
Enable more hints for HLint.

chungyc:
The specific hints that are enabled are `Fuse foldr/map`, `Use :`, `Use camelCase`, `Use concatMap`, `Use maybe`, `Use newtype instead of data`, and `Use replicate`.

These are the last of the hints I will be enabling, which seem uncontroversial enough.  The rest range from those that I am ambivalent about (e.g., `Eta reduce`; point-free style is fine, but I don't feel that we need to strive to be point-free as much as possible) to those I think are a bad idea (e.g., `Use <$`; it's a generally obscure enough operator that isn't used enough in this code base for it to become naturally familiar with).

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -17,5 +17,4 @@
 - ignore: {name: "Use <$>"} # 1 hint
 - ignore: {name: "Use ?~"} # 5 hints
 - ignore: {name: "Use lambda-case"} # 3 hints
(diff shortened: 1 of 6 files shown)


blackgnezdo (approved):
I agree that all the enabled warnings are uncontroversial. Thanks for all these cleanups!

PR #493, opened 2024-10-29
Expand support to Protobuf Editions 2024.

chungyc:
Protobuf Editions 2024 has [no new generic features](https://github.com/protocolbuffers/protobuf/blob/15487415e4077066d5ebdd4357d045cfaa1771b2/src/google/protobuf/descriptor.proto#L964-L1087), so there is no reason not to accept such files for Haskell.  Editions 2024 is useful for specifying the introduction of language-specific features, and non-Haskell features are not useful to us.


blackgnezdo (approved):
Might as well!

PR #498, opened 2024-12-24
Update versions and changelogs.

chungyc:
Only the minor minor version was raised for packages with only version bound changes or other minor non-behavioral changes.  Packages with no changes maintain their versions.  The big version update is for the protoc plugin.


blackgnezdo (approved):
Thanks @chungyc !

PR #509, opened 2025-05-03
Bump to allow ghc-9.10.2 which ships with ghc-prim 0.12.0

blackgnezdo:


chungyc (approved):

ysangkok:
Would be good to release this to Hackage such that proto-lens could get back into Stackage. The current Hackage base bounds do not permit base-4.20.

PR #527, opened 2026-03-23
Correct test-framework dependency

erikd:
Somehow this got screwed up in my last PR. I would have thought that CI would have caught it.


blackgnezdo:
> Somehow this got screwed up in my last PR. I would have thought that CI would have caught it.

Yeah, I'm similarly curious what happened there.

erikd:
I have added a `cabal.project.local` file to my checkout containing:
```
tests: True
```
Maybe you should add that the `cabal.project` file. It just forces the test suites to always be built.

blackgnezdo (approved):

